PR #33, opened 2018-04-15
Add size snapshot

TrySound:
This plugin allows to track size changing with every commit. Also there is useful treeshakability metric which shows size after bundling `import {} from 'lib'`. This allows us to fix treeshakability explicitly without false assumptions.

P.S also I found inconsistency in naming: `dist/react-spring.js` is cjs, `dist/addons.js` is esm.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,0 +1,40 @@
+{
+  "dist/react-spring.js": {
+    "bundled": 65573,
+    "minified": 30885,
+    "gzipped": 8963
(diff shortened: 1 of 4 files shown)


drcmda:
Nice, thanks!

As for addons, that was really weird, i was running into the infamous [class cannot be invoked withoput 'new'](https://github.com/babel/babel/issues/4269) issue (extending classes with es5 turned classes), so i had to bundle the "Animation" class into addons.

PR #38, opened 2018-04-15
Prettify code with close settings

TrySound:
There's a couple of places with too long lines. However I'd like to recommend default 80 print width and 2 tab size.

It's really hard to read the code on my 13 inch mac book where I use half screen for terminal and half for vim.

So if you would like I can change these options.


TrySound:
Done

PR #46, opened 2018-04-16
Refactor object exports with named exports

TrySound:
Refactors modules and won some bytes


drcmda:
Are you planning to amend, or can be merge? Would love to have this in before i attempting to look into injects for the auto-value-fix in the react-dom target.

TrySound:
@drcmda If you agree with everything we can merge this.

drcmda:
Sure looks fine, only `elements` isn't needed in `Spring`.

TrySound:
Good catch

PR #53, opened 2018-04-18
Use createPortal in fix-auto to preserve context for child components

taylorlapeyre:
Possible fix for https://github.com/drcmda/react-spring/issues/52


drcmda:
@taylorlapeyre does it work in react-router?

TrySound:
createPortal should be in react tree. This way shouldn't solve context problem. I think it's not possible to solve the problem out of the tree. `fix` component can be rendered near spring with portal.

PR #56, opened 2018-04-18
Pass Transition onRest through to Spring

### LINKED ISSUE #55: onRest not called when using Transition
`onFrame` works, but `onRest` is never called when using `<Transition />`.

jeef3:
Fixes #55

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,16 +1,16 @@
 {
   "dist/react-spring.es.js": {
-    "bundled": 74973,
-    "minified": 36195,
-    "gzipped": 11298,
+    "bundled": 75051,
+    "minified": 36209,
+    "gzipped": 11302,
     "treeshaked": {
-      "rollup": 28483,
-      "webpack": 29491
+      "rollup": 28460,
+      "webpack": 29468
     }
   },
   "dist/react-spring.umd.js": {
-    "bundled": 88278,
-    "minified": 38262,
+    "bundled": 88356,
+    "minified": 38276,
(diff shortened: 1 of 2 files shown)


jeef3:
💩 _My_ code still works with `4.2.0`. Can't confirm if the bug you're seeing is fixed though.

I'm actually seeing some other issues with using `Transition` and `keys`. It looks like when swapping between two components, that the key assigned is just the function (component). Since I'm interpolating the values myself, this function changes. So `Transition` is unable to match up my elements based on `key`. So I'm seeing elements remaining in the DOM. This is a separate issue though, so I'll see if I can write something up.

drcmda:
@jeef3 is it this you're seeing? https://github.com/drcmda/react-spring/issues/61

If you re-create the function, it is a new component, no matter what, even if the contents remain the same. but if you create it outside, once, and refer to it, transition will be able to recognize it. Otherwise you have to supply your own keys, that would work as well.

jeef3:
@drcmda yeah, I think it might be. I will need to do some refactoring as I think I might be pushing the bounds of what `Transition` is for. As you said, `onRest` shouldn't really apply to `Transition`. The fact that I needed it suggests that I might have a different use case.

In my case, I'm swapping between two components. As far as I can tell, it is not possible to supply keys in this case. I might be just doing it wrong.

What I _think_ I might need is `Spring`, but with the ability to remove it when the animation has finished.

drcmda:
technically that's that transition does, it wraps your component into a spring, but watches lifecycles.

PR #59, opened 2018-04-18
Prefer named exports for normalize color

TrySound:
Saves a few bytes


drcmda:
Could you delete animation/Interpolation ? So weird that it's still there, moved it to targets. Something must be off in my git settings.

PR #60, opened 2018-04-19
WIP: Add TypeScript definition

### LINKED ISSUE #26: Feature Request: Typescript Support
Hi,

This is kind of ground breaking. I think this is er very easy and well thought abstraction for animating stuff based on state. I will definitely take a closer look into this tool :-) 

My biggest problem(which is kind of small) at the moment is that there aren't any typings to find for this package yet. I would really love to see some typings for this libary as it would make the whole API easier to get into and make it way easier to use in Typescript projects :-)

Cryrivers:
Closes #26

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,62 @@
+declare module 'react-spring' {
+    import { PureComponent, ReactNode, ComponentClass } from 'react';
+    type SpringConfig = {
+      tension: number;
+      friction: number;
+    };
+    type SpringProps<S extends object, DS extends object = {}> = {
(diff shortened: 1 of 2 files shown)


Cryrivers:
@drcmda still WIP 😢 but should be usable for `Spring`, `animated` and `interpolate`.

drcmda:
Oops, 😓could you make another? I thought this is only targeting Spring for now ...

Thanks for putting work into this!

Cryrivers:
@drcmda np,i'll make a PR soon.

PR #63, opened 2018-04-19
[docs] fixed two links

jlarmstrongiv:
Small fix for two links in the documentation.  Also, capitalized README.md since all the other docs were capitalized.  

---

One quick question with the docs.  I was reading up on the native and animated component.  The usage suggests `animated.div` and I noticed it works with other html elements like `animated.li`.  

I tried to get it to work with a react component `const MyDiv = props => <div>{props.children}</div>
`, but `animated.MyDiv` didn’t work.  ([https://codesandbox.io/s/l419mkpn69](https://codesandbox.io/s/l419mkpn69))  

I assume I cannot use animated with functional or class components?

### DIFF
--- a/API-OVERVIEW.md
+++ b/API-OVERVIEW.md
@@ -3,7 +3,7 @@
 
 * [Springs and basic interpolation](#springs-and-basic-interpolation)
 * [Render props](#render-props)
-* [Native rendering and interpolation](#native-rendering-and-interpolation)
+* [Native rendering and interpolation](#native-rendering-and-interpolation-demo)
 * [Imperative Api](#imperative-api)
 * [Transitions](#transitions)
 * [Parallax and page transitions](#parallax-and-page-transitions)
@@ -98,7 +98,7 @@ const App = ({ children }) => {
   const hover = () => spring(animation, { to: '#c23369' }).start()
   const unhover = () => spring(animation, { to: '#28d79f' }).start()
   return (
-    <animated.div 
+    <animated.div
       style={{ background: animation }}
       onMouseOver={hover}
       onMouseOut={unhover} />
--- a/readme.md
+++ b/readme.md
@@ -219,7 +219,7 @@ import { Spring, animated } from 'react-spring'
 </Spring>
 ```
 
-More about native rendering and interpolation [here](https://github.com/drcmda/react-spring/edit/master/API-OVERVIEW.md#native-rendering-and-interpolation).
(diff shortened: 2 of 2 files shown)


drcmda:
@jlarmstrongiv 

```jsx
const MyDiv = ({ height, ...rest }) => <animated.div style={{ height }} {...rest}/>

          <Spring native from={{ height: 0, opacity: 0 }} to={{ height: toggle ? 'auto' : 0, opacity: 1 }}>
            {({ height, opacity }) => (
              <MyDiv className="item" height={height}>
                {text.map(t => <animated.p style={{ opacity }}>{t}</animated.p>)}
              </MyDiv>
```

You're animating 'auto', so you want to apply it on the parent element which contains the items. In your case that would be the div inside MyDiv. For kicks, i extended the paragraph from animated as well, it would fade in like this ...

jlarmstrongiv:
Ahh alrighty, thank you for the example!  I was trying to work out the best way to use the native animated component with the [styled-components](https://www.styled-components.com/docs/basics#adapting-based-on-props) library.  Going through props (or redux, etc.) seems like the way to go 🙂

drcmda:
If you have a small demo for styled components, would be very interested in it. I haven't had the time to look into how the two libs behave.

jlarmstrongiv:
Sure, it looks like style-components can be used with the animated component through the [`styled` method](https://www.styled-components.com/docs/basics#styling-any-components).

```jsx
import { render } from 'react-dom';

import React from 'react';
import styled from 'styled-components';
import { Spring, animated, config } from 'react-spring'

const MyParagraph = ({ className, style, children }) => (
  <animated.p className={className} style={style}>
    {children}
  </animated.p>
);

const MyStyledParagraph = styled(MyParagraph) `
  color: palevioletred;
  font-weight: bold;
`;

const App = props => (
  <div>
    <Spring native from={{ opacity: 0 }} to={{ opacity: 1 }} config={config.slow}>
      {style => <MyStyledParagraph style={style}>i will fade in</MyStyledParagraph>}
    </Spring>
  </div>
);

render(<App />, document.getElementById('root'));
```

Here it is on [codesandbox](https://codesandbox.io/s/4812v8mmo4) 🎉 (though it’s hard to see the fade with the page load, even with the config.slow)

drcmda:
`config={{ tension: 10, friction: 50 }}` 😉

Thanks for the sandbox!

Issue #68, opened 2018-04-24
Spring config for individual value


crimx:
In react-motion I can use different `spring` wrappers for different values, which comes quite handy for controlling values for other components to animate. It seems that react-spring can only set `config` to the component. Will there be support for this? Or is there a better/idiomatic way to achieve this with react-spring?

My use case: I have an icon that may translate and scale. When it mounts, it translates immediately to the position and scales wobbly to `1`; otherwise it just translates gently to the cursor position.

Here is a simple example with react-motion: https://jsfiddle.net/straybugs/0cxobcut/

drcmda:
@crimx makes sense, i'll add it. I wanted to remove all the parts that i felt like were rarely used but would stretch the api either way. 

I could make the config option accept either an object or a function. The function would receive the key. While i'm at it maybe it's a good idea to do the same for `immediate`.

Would that solve your issue?

crimx:
@drcmda That'll do. I'd say this approach is better. Much cleaner.

drcmda:
@crimx you can try it now: @4.2.1-beta.1

```jsx
<Spring
    config={name => name === 'y' ? specialConfig : normalConfig}
    immediate={name => name === 'x'}
    ...
```

-> don't animate "x" and give "y" a special config

crimx:
Wow that's super fast! I'll try it.

crimx:
Looks good. Here is the example.

[![Edit 128n84yqj](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/128n84yqj)

crimx:
@drcmda Thanks for the tips! This is actually the reason why I decided to switch to react-spring! I didn't use it here because in my real use case, the icon is injected to the page, so I need to get around react to set important styles.

PR #69, opened 2018-04-25
Update Spring types

crimx:
Updated Spring types. Didn't touch other components since I haven't used them yet.


crimx:
#68 related. I also added `impl` and `inject` to Spring props and made `children` optional.

PR #76, opened 2018-05-03
Move fix-auto from portal-based to in-place hidden measuring

a-type:
### Summary

The portal-based solution resulted in inaccurate width calculations on the user's content when the content did not have a constrained width value and contained wrapping text.

This solution uses a 0-height rendering of the user's content in-place, so that all layout constraints stay consistent with intended usage. I used a little trick to get the total height based on the `scrollHeight` on the 0-height measuring element.

I'm not convinced this will work with all layouts, since we're rendering this intermediate div, which breaks any layout properties which the user's content might be getting from its parent. Still, I think this solution may fit a wider variety of use cases than the current one.

### DIFF
--- a/src/animated/targets/react-dom/fix-auto.js
+++ b/src/animated/targets/react-dom/fix-auto.js
@@ -29,6 +29,7 @@ export default function fixAuto(spring, props) {
 
   return (
     <div
+      style={{ visibility: 'hidden', position: 'absolute' }}
       ref={ref => {
         if (ref) {
           // Once it's rendered out, fetch bounds
--- a/src/animated/targets/react-dom/fix-auto.js
+++ b/src/animated/targets/react-dom/fix-auto.js
@@ -27,18 +27,12 @@ export default function fixAuto(spring, props) {
     ? allProps.reduce(convert, forward)
     : { ...from, ...to, ...forward }
 
-  // Setting height and scroll properties so that the measuring div will be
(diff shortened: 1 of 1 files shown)


a-type:
Yeah, I've noticed that this solution is not exactly perfect... if wrapping text just barely fills the entire space, it seems like there's still the possibility of the height being calculated as if there's an extra line (probably the measured version wrapped for some reason).

drcmda:
@a-type  Made some tests as well and saw that in some situations the scrollbar would also push content inwards a little. If the previous version (just rendering, no portal) didn't cause noticeable flicker, maybe we run with this one for a while until we figure out something better?

a-type:
@drcmda Good call on the scrollbar! I bet I could just hide the scrollbar entirely to avoid this and it may fix it.

a-type:
Yeah, I can't find a good cross-browser solution for hiding the scroll bar. I'll go for the simple route.

a-type:
Just did a bit more research. Apparently the classic trick is to do `visibility: hidden; position: absolute`. Setting position to absolute doesn't actually change the element bounds as long as you don't set `left/right/top/bottom`, so it seems valid. I tried it out on my test cases and it seems to work!

a-type:
Nevermind about the visibility thing. I found a few cases where it reports incorrect width and throws off the calculations again.

PR #84, opened 2018-05-09
First try at adding missing type annotations.

NulledGravity:
I don't have much experience with TS, haven't had the chance to test the annotations, let me know if there's something that doesn't work.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 declare module 'react-spring' {
-  import { PureComponent, ReactNode, ComponentClass } from 'react';
+  import { Component, PureComponent, ReactNode, ComponentClass } from 'react';
 
   export type SpringConfig = {
     tension: number;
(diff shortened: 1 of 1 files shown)


NulledGravity:
I can't understand why the travis checks have failed. I've modified only `index.d.ts`, and after reading the logs it didn't seem that it's a problem with my changes but with some dependencies of `react-spring`.

drcmda:
@NulledGravity not sure what's travises problem, but it's not your fault. I get the same error, though the local build seems to be fine. Is your PR complete?

NulledGravity:
@drcmda I've added the types following [https://github.com/drcmda/react-spring/blob/master/API.md](https://github.com/drcmda/react-spring/blob/master/API.md). I'm relatively new in TS and haven't had the chance to test if all the types are correct. I'd like to get some advice/help on how to complete the annotations.

Also I still need to comment all the types.

Any help welcome :)

drcmda:
Great! Could you make a last check and see if it corresponds to the actual proptypes in the js files? API.md could be slightly outdated since there were some changes recently, for instance to the way config and immediate work, they can be functions now that return individual results. If you think the PR is done i'll merge - i myself have zero knowledge sadly when it comes to TS.

index.d.ts
```diff
@@ -95,4 +95,122 @@ declare module 'react-spring' {
...
+    ) => ReactNode | Array<(params: DS & S) => ReactNode>;
+
+    render?: (
+      params: DS & S,
+    ) => ReactNode | Array<(params: DS & S) => ReactNode>;
+  };
+
+  export class Transition<
+    S extends object,
+    DS extends object,
```
jariz: I'm getting TS1009 no trailing spaces allowed on the latest version of TS, could you please remove them?

jariz:
@drcmda yep. only spotted this because I was using it in my exact usecase, but def gonna need some tests at some point down the road to double check if they're all correct.

PR #88, opened 2018-05-16
🔥 Remove console log

epilande:
I've pulled down the latest of `react-spring@5.1.8` and started seeing "inject done" in the console. This PR removes that `console.log`.

BTW thanks for the great lib. 😄


drcmda:
@epilande it was a left over, i already removed it today. : -)

PR #107, opened 2018-06-05
fix addon import path in docs

sourcesoft:


drcmda:
What's wrong with the default route? It points to the es-module: https://unpkg.com/react-spring@5.3.8/dist/addons.js

sourcesoft:
I couldn't import like it was documented and then took a look at https://github.com/drcmda/react-spring/blob/master/rollup.config.js#L39 and https://codesandbox.io/embed/q9lozyymr9 so I added .cjs to the end.
If it's just me then it's a false alarm.

drcmda:
Yeah, i had the same problem with csb, maybe their bundler is not equipped to handle esm? I opened an issue, let's wait and see. In the demos i load addons as well using the em module (https://github.com/drcmda/react-spring/blob/master/examples/demos/timing/index.js) and it works in a generic webpack setup.

sourcesoft:
Thanks. I can import from `src` folder too but the problem seems to be the packaged npm bundle in the `dist` folder. I'm using Next.js btw and here's my `dist` folder in `node_modules`. Not sure how it works with rollup but there are no bundles ending with just `.js` here:
screen shot 2018-06-05 at 4 41 33 pm

drcmda:
In a previous version the es exports had ".es" in the name, but now they're just the default, only cjs and umd are marked. If next uses webpack 4 it should be able to deal with modules, i think only the older webpack variants couldn't do it.

drcmda:
@sourcesoft it was a bug in codesandbox, Ives said it's fixed now.

PR #123, opened 2018-06-19
Add onStart callback

Cryrivers:
Hi, I'm adding `onStart` callback because I'd like to know the time that animation starts to animate, which allows me to optimize the rendering performance.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -10,8 +10,8 @@
     "gzipped": 2141
   },
   "dist/web.umd.js": {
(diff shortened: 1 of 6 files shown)


drcmda:
@Cryrivers would merge, but can you move it down into the components start function? entry.start wouldn't work out, because if you have 100 props, it would fire onStart 100 times on each start.

Cryrivers:
i didn't see that `start()` before 😂. now fixed.

PR #127, opened 2018-06-21
Add SpringRendererFunc typing

nelsliu9121:
I recently encountered that TypeScript is throwing typing error when I use `<Transition />` because `children` type does not match.
When I inspected the typing file in the repo, I found out that the definition for `render` and `children` could cause some misunderstanding.

Therefore I offer a fix to this, take a look at the changes.


drcmda:
@nelsliu9121 nice, would you add a proper key definition as well? https://github.com/drcmda/react-spring/issues/126

PR #146, opened 2018-07-11
Fix keyframes types and use interface so user can extend

thomhos:
I noticed that the Keyframes typings don't seem to be in sync with the latest version anymore, so this should fix that.

Also changed the type defs to use interfaces as that actually allows people to extend them. For instance ParallaxLayer should also accept a `style` prop. You should be able to easily extend that as end-user. 

Also, I have no idea what all these S, DS generics are. They seem to be used all over the place even when they're not being used in the actual type / interface. See `ParallaxLayerProps` for instance. Maybe the original author knows why thats there?

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "dist/addons.umd.js": {
-    "bundled": 15720,
-    "minified": 5853,
-    "gzipped": 2141
+    "bundled": 16003,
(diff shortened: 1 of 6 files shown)


drcmda:
@thomhos hard to say, they're community managed. Is this good to merge?

thomhos:
Yes

PR #157, opened 2018-07-23
Update Transition type definition

lasseborly:
This PR reflects some of the thought put out in https://github.com/drcmda/react-spring/issues/26#issuecomment-406741416.

From reading some of the recent PR's I can see that having a literal boolean type on `Transition` has been discussed in #84 but never seemed to have gotten implemented.
The TS compiler throws an error if a literal boolean false type is being substituted for this PR's basic boolean type. A simple type discrepancy. Therefore `boolean` is used instead.

I'll modify as needed.

Thanks for a great lib @drcmda 👍

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,14 +15,14 @@ import { Spring, ... } from 'react-spring/dist/universal'
 
 # Table of Contents 👇
 
-* [What is it?](#what-is-it-)
-* [Why do we need yet another?](#why-do-we-need-yet-another-)
-* [Links](#links-)
-* [Basic overview](#basic-overview-)
-* [Interpolation](#interpolation-)
-* [Render props](#render-props-)
-* [Native rendering](#native-rendering-)
-* [React-native and other targets](#react-native-and-other-targets-)
+- [What is it?](#what-is-it-)
+- [Why do we need yet another?](#why-do-we-need-yet-another-)
+- [Links](#links-)
+- [Basic overview](#basic-overview-)
+- [Interpolation](#interpolation-)
+- [Render props](#render-props-)
+- [Native rendering](#native-rendering-)
+- [React-native and other targets](#react-native-and-other-targets-)
 
 # What is it? 🤔
 
@@ -67,7 +67,6 @@ react-spring is a cooked down fork of Christopher Chedeau's [animated](https://g
 react-spring builds upon animated's foundation, making it leaner and more flexible. It inherits react-motions declarative api and goes to great lengths to simplify it. It has lots of useful primitives, can interpolate mostly everything and last but not least, can animate by committing directly to the dom instead of re-rendering a component frame-by-frame.
 
 For a more detailed explanation read [Why React needed yet another animation library](https://medium.com/@drcmda/why-react-needed-yet-another-animation-library-introducing-react-spring-8212e424c5ce).
-
 
 # Links 🔗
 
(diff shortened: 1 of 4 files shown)


src/Transition.js
```diff
@@ -52,10 +52,12 @@ export default class Transition extends React.PureComponent {
     children: PropTypes.oneOfType([
       PropTypes.arrayOf(PropTypes.func),
       PropTypes.func,
+      PropTypes.bool,
```
drcmda: @lasseborly  `{visible && fn}` isn't a boolean, though. It either results in a function of undefined.

src/Transition.js
```diff
@@ -52,10 +52,12 @@ export default class Transition extends React.PureComponent {
...
     ]),
     render: PropTypes.oneOfType([
       PropTypes.arrayOf(PropTypes.func),
       PropTypes.func,
+      PropTypes.bool,
```
drcmda: same here.

lasseborly:
@drcmda I see your point. Will correct.

lasseborly:
@drcmda corrected!

readme.md
```diff
@@ -127,11 +126,22 @@ Given a single child instead of a list you can toggle between two components.
 import { Transition } from 'react-spring'
 
 <Transition from={{ opacity: 0 }} enter={{ opacity: 1 }} leave={{ opacity: 0 }}>
-    {toggle ? ComponentA : ComponentB}
+    {toggle
+        ? styles => <ComponentA style={styles} />
+        : styles => <ComponentB style={styles} />
```
epilande: This will definitely clear up some confusion 👍

lasseborly:
@drcmda is there anything else I can do to make this go through?
Would love to get off of my own fork and correct my dependency.

drcmda:
@lasseborly i'll merge and release today. wasn't with a computer over the last week, we made a small trip to rome :-)

lasseborly:
Sounds lovely @drcmda !

drcmda:
@lasseborly 

Just went through, boolean still seems to be part of children and render

```jsx
  children?:
    | SpringRendererFunc<S, DS>
    | Array<SpringRendererFunc<S, DS>>
    | boolean

  render?:
    | SpringRendererFunc<S, DS>
    | Array<SpringRendererFunc<S, DS>>
    | boolean
```

Could i see a snippet of code that depends on your fork? Because it seems strange, children/render shouldn't take booleans, only functions and ternaries that yield functions, or undefined.

lasseborly:
@drcmda so when we do the single component transition TS interprets that as children/render being a boolean if `toggle` is falsy.

I made a [codesandbox](https://codesandbox.io/s/8krrok63z2) so you can see the problem for yourself.
This problem get's solved if we add the `boolean` possibility to the TS type def.
This is also why I made a reference to a previous PR #84 which tried to solve this by adding a literal type. The literal type is not the correct solution though, because TS still interprets the expressions to have the possibility of being a `boolean`.

lasseborly:
I can see the argument for not having the `boolean` prop-type on the JS side of things, since as you pointed out this evaluates to `undefined` and not specifically `false` and therefore only a falsy type not an actual `boolean`.

drcmda:
Just checking if it's correct - i have yet to dig into TS. If it helps TS knowing undefined i guess it's fine. I'll run some errands and release later this day.

PR #163, opened 2018-07-24
Update type definitions

PrecociouslyDigital:
State is optional on keyframes when you have a default animation; delay was missing on SpringProps

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -105,7 +105,7 @@ export const config: {
 }
 
 export class Spring<S extends object, DS extends object> extends PureComponent<
-  SpringProps<S, DS>
+  SpringProps<S, DS> & S
 > {}
 
 export function interpolate(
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -113,10 +117,10 @@ export const animated: {
   [Tag in keyof JSX.IntrinsicElements]: ComponentClass<
     JSX.IntrinsicElements[Tag]
   >
-}
+};
 
-type TransitionKeyProps = string | number
-type TransitionItemProps = string | number | object
+type TransitionKeyProps = string | number;
+type TransitionItemProps = string | number | object;
```
lasseborly: Intentionally added lonely semi-colons? 😊

PrecociouslyDigital: Oops; My linter is having a fit with the code styles and might have accidentally added them. Personally I think the standard should be semi-colons, but I thought it was better to be consistent

lasseborly: Yeah I agree @PrecociouslyDigital. Cool!

PR #164, opened 2018-07-24
Update API.md

PrecociouslyDigital:
Hold no longer exists on Spring

### DIFF
--- a/API.md
+++ b/API.md
@@ -53,8 +53,6 @@ class Spring extends React.PureComponent {
     render: PropTypes.func,
     // Prevents animation if true, you can also pass individual keys
     immediate: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
-    // Won't start animations, so they can be controlled from outside
-    hold: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
     // Spring config ({ tension, friction, ... } or a function receiving a name)
     config: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
     // Animation start delay, optional


zachasme:
Why was this removed? Is there a new way to achieve the effect?

drcmda:
I didn't know what it was tbh. Internals changed in 6.0, everything is controlled by a class called `Controller`, see: http://react-spring.surge.sh/manual, spring basically wraps around it. In the original animated library AnimatedValue and Animated.spring would control animations, which was problematic for performance reasons and it caused unwanted complexity. 

From looking at old code (seems like this was removed long time ago actually ... 🤔)

```jsx
        // Skip held animations
        if (hold && (hold === true || hold.indexOf(name) !== -1))
          return onFinish()

        AnimatedController(entry.animation, { toValue, ...config }, impl).start(
          props => props.finished && onFinish()
        )
```

to me it looks like it is the same as `immediate` but without jumping to end-state. I wonder though, wouldn't it be the same if you just don't set values?

Could i see an example of the effect in action?

drcmda:
@PrecociouslyDigital API.md is a relic, could you remove the whole thing? Props are all extracted automatically from PropTypes in the official documentation now.

zachasme:
Discussion was in #48. I'll have to look into the new internals, it's been a while since I had the hold prop merged, perhaps it's no longer necessary.

drcmda:
Now i remember, thanks! I think this actually works out of the box now. The demos (especially the one with the red good drops): http://react-spring.surge.sh/manual circumvent the render phase on mouse input. This is because you can update a controller anywhere. This is very easy to do with hooks (and in that example you see how you can even re-use AnimatedValues in further springs. But theoretically this would also work with a generic <Spring />:

```jsx
componentDidMount() {
  // This would animate views, but it won't cause a new render pass
  this.spring.controller.update({ ...newprops }, true)
}
render() {
  return (
    <Spring ref={this.spring} ... >
      {props => <animated.div style={props} />
```

If there's something missing or if we need to make this easier we could discuss this in a separate issue.

drcmda:
Here's a proof of concept: https://codesandbox.io/s/534v0v8pyx

It's not really as straight forward as hooks or imperative controller because it needs (to: {...}) but other than that it works.

PrecociouslyDigital:
@drcmda I added a pull request to remove the bit, but it has a trivial merge conflict I will resolve when I get home

PR #185, opened 2018-08-17
Feature: Add ref forwarding to animated components

psachs21:
Resolves https://github.com/drcmda/react-spring/issues/184


psachs21:
Unfortunately, tests are failing due to https://github.com/airbnb/enzyme/issues/1604.

Really bummed about it but it looks like it should be resolved in 3.50

drcmda:
@psachs21 looked into it some more and it's fine, let's use it!

PR #193, opened 2018-08-22
Remove redundant argument passed to parseFloat

Andarist:


src/targets/shared/normalizeColors.js
```diff
@@ -183,7 +183,7 @@ function parse255(str) {
 
 function parse360(str) {
   const int = parseFloat(str)
-  return ((int % 360 + 360) % 360) / 360
+  return (((int % 360) + 360) % 360) / 360
```
Andarist: prettier

PR #201, opened 2018-08-23
Tweak animated

Andarist:


src/targets/web/index.js
```diff
@@ -151,12 +151,10 @@ const domElements = [
   'tspan',
 ]
 
-const elements = domElements.reduce(
-  (acc, element) => ({ ...acc, [element]: animated(element) }),
-  {}
-)
-
-Object.assign(animated, elements)
+const extendedAnimated = domElements.reduce((acc, element) => {
```
drcmda: I think this could this be expressed as:

```jsx
domElements.forEach(element => animated[element] = animated(element))
```

Andarist: sure, the "problem" I'm trying to solve is to always output a variable whenever possible (`forEach` wont do that), that way minifiers can track more easily what's used and what's not

if you have
```js
const a = []
const b = /*__PURE__*/ ['a','b','c'].reduce((acc, el) => {
  acc.push(el)
  return acc
}, a)

export { b }
```
and `b` stay unused then the minifier will remove its declaration, but because it was created with function call (`.reduce`) it will only know that it can remove the call itself if its result stay unused (and it's no longer used in this example) thanks to `#__PURE__` annotation and once it removes this function call there will be no longer any references to `a` so it will be able to remove it too

the "trick" is to always create a "linked" variables, even if u do mutations, because in this example:
```js
const a = []
['a','b','c'].forEach(el => {
  a.push(el)
})
export { a }
```
it would never be able to remove `a` because its always used by the forEach loop, and if we would add `#__PURE__` annotation to the forEach call it would ALWAYS remove it, because its result is unused (there is ofc no result here, but uglify doesnt know that)

PR #232, opened 2018-09-17
Fix onRest & animated typing

jigsawye:
I think the `onRest` callback will receive the animated value, but I haven't any experience of TypeScript, so please help me to review it, thanks!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -121,7 +121,7 @@ export const animated: {
   [Tag in keyof JSX.IntrinsicElements]: ComponentClass<
     JSX.IntrinsicElements[Tag]
   >
-};
+}
 
 type TransitionKeyProps = string | number
 type TransitionItemProps = string | number | object
--- a/index.d.ts
+++ b/index.d.ts
@@ -116,10 +116,12 @@ export function interpolate(
 ): any
 
 export const animated: {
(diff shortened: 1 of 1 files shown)


drcmda:
That would be tough since i don't use typings myself. Anyone else?

jigsawye:
The `animated` also need to fixed, it is defined as object, so I can't use it as a function:

```ts
export const animated: {
  [Tag in keyof JSX.IntrinsicElements]: ComponentClass<
    JSX.IntrinsicElements[Tag]
  >
}
```

```js
src/Modal/BaseModal.tsx:36:30 - error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '{ a: ComponentClass<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, any>; abbr: ComponentClass<DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>, any>; ... 167 more ...; view: ComponentClass<...>; }' has no compatible call signatures.

36 const AnimatedModalContent = animated(ModalContent);
```

drcmda:
@jigsawye  Is this good to go in?

jigsawye:
@drcmda Yes, the fixed typing working correct in my project, but I don't know whether the changes is best practice.

PR #250, opened 2018-10-03
Forward element key when calculating height auto

einarlove:
Fixes some part of https://github.com/drcmda/react-spring/issues/249

### DIFF
--- a/src/targets/web/fix-auto.js
+++ b/src/targets/web/fix-auto.js
@@ -23,6 +23,7 @@ export default function fixAuto(props, callback) {
   // Return v.dom with injected ref
   return (
     <element.type
+      key={element.key}
       {...element.props}
       style={{ ...elementStyles, position: 'absolute', visibility: 'hidden' }}
       ref={ref => {


einarlove:
Would it be possible to release a patch version to NPM? I'm quite dependent on this fix and I can't install from this commit directly via `github:drcmda/react-spring#ca1e401ea1e33237042b8d2787f50f703ee37911` because your build script is dependent on YARN to work.

drcmda:
no problem, i am trying to get one last fix in

drcmda:
@einarlove published @5.9.1-beta.1

It contains bugfixes for your transition case as well. https://codesandbox.io/embed/mmvk3p3xw8

einarlove:
I recorded the inconsistent behavior of my Keyframes within a Transition which also animates height:auto.

bug-report

*The gif shows how the spinner around the buttons stops at random points. When used outside the Transition it correctly animates from 0 to 100% of the circle with Easing.linear*

einarlove:
I'll try to post an example where it fails. Might time some time since it's hard to decouple my components from my project atm.

PR #257, opened 2018-10-09
example tests for animated library

chanceeakin:


drcmda:
@chanceeakin very nice thanks for these tests! 😊

drcmda:
PS. i've refactored a little - all going into the 6.x-experimental branch right now. Only in case that you have something cooking. Deleted AnimatedTracking, the controller and SpringAnimation.

chanceeakin:
@drcmda: you're welcome! I'll take a look this evening (I'm out of Austin) and kick you some more stuff. Do we have the inheritance tree for `src/animated/` written down anywhere?

drcmda:
The best i've found was:

```jsx
/**
 * Animated works by building a directed acyclic graph of dependencies
 * transparently when you render your Animated components.
 *
 *               new Animated.Value(0)
 *     .interpolate()        .interpolate()    new Animated.Value(1)
 *         opacity               translateY      scale
 *          style                         transform
 *         View#234                         style
 *                                         View#123
 *
 * A) Top Down phase
 * When an Animated.Value is updated, we recursively go down through this
 * graph in order to find leaf nodes: the views that we flag as needing
 * an update.
 *
 * B) Bottom Up phase
 * When a view is flagged as needing an update, we recursively go back up
 * in order to build the new value that it needs. The reason why we need
 * this two-phases process is to deal with composite props such as
 * transform which can receive values from multiple parents.
 */
```

It goes through AnimatedController(AnimatedValue:v) -> v.update -> flush -> AnimatedProps -> AnimatedStyles or something. I already ripped most of it out but the overall structure is still standing, though i still don't know what it has to be so nested.

PR #259, opened 2018-10-10
reorganized testing folders and added more snapshots

chanceeakin:
Changed around the directory structure to keep tests closer to the source files. Ditch it if you don't like it. 😄


drcmda:
@chanceeakin could you take another look? i've been so busy with 6.0 that i was swamped a little. some minor things have changed, api as well - types and tests are screwed atm. the example suite works at least.

chanceeakin:
Absolutely. I'll give it a run sometime over the weekend! Congratulations on 6.0! 😄

chanceeakin:
Give it a look over and lmk if this is a direction you'd like to proceed. :)

drcmda:
Thanks, looks great!

drcmda:
Btw, do you know what role does storybook plays in this? I saw that some tests are using components out of the storybook folder.

chanceeakin:
Earlier tests were using storybook components as mocks, as far as I could tell. Didn't want to disturb that pattern if it needed to stay.

PR #263, opened 2018-10-13
testing keyframe internals

chanceeakin:


drcmda:
Keyframes.Transition doesn't exist any longer, Transition itself now uses Keyframes internally. The mount status looks unusual to me, wouldn't doing it with setState cause sideeffects since setState is async? Calling it on componentWillUnmount for instance will potentially let callbacks that fire in this moment still allow to work because the mounted=false flag will either come never (not sure if React fires setState after cWU or later.

chanceeakin:
Haha, you're so right. I'll nuke this one. This was a PR for 5.x

PR #270, opened 2018-10-21
Properly infer generic typings for Spring/Transition/Trail

malwilley:
Currently in `<Spring />`, `<Transition />`, and `<Trail />`, the child function parameter types cannot be inferred. As an example:

```jsx
<Spring from={{ opacity: 0 }} to={{ opacity: 1 }}>
	{(styles) => <div style={styles} />}
</Spring>
```

Here `styles` would be of type `any`. With the proposed changes however, `styles` would be of type `{ opacity: number, [key: string]: any }`. The `[key: string]: any` is an unfortunate side effect of needing to support passthrough props. I don't see passthrough props documented anywhere but they are used in [this demo](https://codesandbox.io/embed/oln44nx8xq) (`toggle`) so I made sure to support it in the types. If that isn't actually part of the API I can take out the `& WeakObject`'s. and make this a bit cleaner.

For `Transition` I made the assumption that the child function is called with the intersection of `initial`, `from`, `enter`, `leave`, and `update` if these props were supplied with different types. As another example:

```jsx
<Transition
	from={{ opacity: 0 }}
	enter={{ opacity: 1, height: 100 }}
	leave={{ width: 0 }}>
	{(styles) => <div style={styles} />}
</Transition>
```

Here `styles` would be of type:

```ts
{
	height: number,
	opacity: number,
	width: number
	[key: string]: any,
}
```

Reference https://github.com/drcmda/react-spring/issues/26 again since this improves more on the Typescript support.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,13 +21,11 @@ export interface SpringConfig {
   easing?: SpringEasingFunc
 }
 
-type SpringRendererFunc<S extends object, DS extends object = {}> = (
-  params: DS & S
-) => ReactNode
+type SpringRendererFunc<DS extends object = {}> = (params: DS) => ReactNode
 
-interface SpringProps<S extends object, DS extends object = {}> {
+interface SpringProps<DS extends object = {}> {
   /**
-   * Spring config ({ tension, friction })
+   * Spring config, or for individual keys: fn(key => config)
    * @default config.default
    */
   config?: SpringConfig | ((key: string) => SpringConfig)
@@ -40,7 +38,7 @@ interface SpringProps<S extends object, DS extends object = {}> {
    * Base styles
    * @default {}
    */
-  from?: DS
+  from?: Partial<DS>
   /**
    * Animates to...
    * @default {}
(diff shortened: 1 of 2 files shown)


drcmda:
Could we hold this for the major that comes out in the next days? If you like, take a look in the 6.x-experimental branch. There were some small changes that would conflict with this i think. children/render changes a little. Children only takes a single child, even for transitions and trails. Render is out since i think no one ever used that and it just brought more complexity.

malwilley:
Good call out, probably not worth porting over these changes if 6.x is being released soon. I could either make a PR against the 6.x branch or let this one sit until the new release is out. Whichever you prefer.

On another note, are the passthrough props (not sure I'm using the correct term here) that I mentioned something you want to allow typings for? I'm personally not a fan of them since I'm not aware of any way to infer types safely (hence the `[key: string]: any`), but if that's a popular feature of the library then obviously they'd have to stay.

drcmda:
Right, the pass through props were another weird thing, but i'd like to take it out. This stuff only creates headaches and can't be typed.

drcmda:
Should i close this in the meantime?

malwilley:
Oh, didn't realize 6.0 was released! I can fix this up tonight and get it ready for review again.

malwilley:
Sorry for the delay! Updated for the 6.0 API and also fixed up some of the item/children types for Transition/Trail. Ready for review @drcmda

drcmda:
Nice, thanks a lot!

PR #271, opened 2018-10-23
Add ParallaxLayer position update hook

changalberto:
@drcmda This hook is intended to allow users to tap into the current state of a layer position so that they can animate elements according to those values. Please let me know if there could be a better way to handle this.

### DIFF
--- a/API-OVERVIEW.md
+++ b/API-OVERVIEW.md
@@ -307,7 +307,7 @@ const Container = Keyframes.create(MyOwnPrimitive)({ ... })
 import { Parallax, ParallaxLayer } from 'react-spring'
 
 <Parallax pages={3} scrolling={false} horizontal ref={ref => this.parallax = ref}>
-    <ParallaxLayer offset={0} speed={0.5}>
+    <ParallaxLayer offset={0} speed={0.5} onPositionChange={(to, height) => {}}>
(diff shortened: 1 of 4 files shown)


drcmda:
Looks good to me, did you close by accident? I would just ask you to hold this until 6 is out. I had to refactor parallax a little and it's now under addons. You can look into the 6.x-eperimental branch to see it.

changalberto:
No, I closed the PR on purpose. My updates wasn’t very helpful in trying to achieve the results I want… On that note, maybe you can point me in the right direction.

I’m trying to make an object/element animate within the Parallax Layer based on its Layer “revealed” percentage. In other words, I want the animations frames to be based on scrolling.
Using the Vertical Parallax example given in the repo, ideally, the code should looks something like this:

https://monosnap.com/direct/cqsaGWXSgD7dYWqps29PiCCqQ2DcyJ <https://monosnap.com/direct/cqsaGWXSgD7dYWqps29PiCCqQ2DcyJ>

Please let me know if this is do-able and what steps I should take to make this enhancement.

Note: I’m not a React Expert nor am I very familiar with React Spring. I just started dabbling with this Animation Framework (I’m loving it so far. Thanks for making this happen).

PR #272, opened 2018-10-24
Quick typo fix in Spring.mdx

ooloth:


ooloth:
Hey, I'm having some trouble rebasing (I'm still a bit new with the PR workflow...), so I'll just resubmit instead for simplicity.

Issue #280, opened 2018-10-25
React / React DOM are relative pathed


rmartin:
React and React DOM were committed as a relative path dependency in the latest build, which prevents from compiling locally without this reference. I'll open a PR to move this back to the latest release that was announced as I believe that was the intent with 'react-future' references.

drcmda:
Ah yes, this was from messing around locally.

PR #281, opened 2018-10-25
280 - address issues with React / React DOM as relative path.

rmartin:
Addresses relative path issue in release 6.0.0. #280


rmartin:
This was addressed with a commit directly to master.

PR #282, opened 2018-10-25
248 ie11 browser support

rmartin:
Addresses issues with IE11 and others that don't support ES2017 object entries. #279 . Also includes code fix for pr #281


rmartin:
@drcmda this should now be updated and working as expected. I included the size snapshot in this commit as well as this changed with the additional package. It appears you're committing that as well but let me know if that needs to be removed for any reason.

drcmda:
Run some tests with it, i am uncertain now if this is the right thing to do. Entries is used in many places and some performance-critical. It would add half a kb and slow down execution. Wouldn't it be way simpler to just add a basic polyfill if you want this to run on IE11? Otherwise 99.9% of the our users would be hit with slight regression for the 0.1% that are still using this browser. 

Considering how easy it is to polyfill it, what do you think?

PR #321, opened 2018-11-12
implementing transition hook. WIP

### LINKED ISSUE #320: Implement Transitions Hook

ruggedy:
closes  #320

### DIFF
--- a/src/addons/KeyframesHook.js
+++ b/src/addons/KeyframesHook.js
@@ -1,0 +1,113 @@
+import React from 'react'
+import { useSpring } from './SpringHook'
+import { useTrail } from './TrailHook'
+
+export function parseKeyframedUpdate (slots, f, setNext) {
+  if (Array.isArray(slots)) {
+    let q = Promise.resolve()
+    for (let i = 0; i < slots.length; i++) {
+      const slot = slots[i]
+      const last = i === slots.length - 1
+      q = q.then(() => setNext(f(slot), last))
+    }
+  } else if (typeof slots === 'function') {
+    slots((animatedProps, last = false) => setNext(f(animatedProps), last))
+  } else {
+    const last = true
+    setNext(f(slots), last)
+  }
+}
+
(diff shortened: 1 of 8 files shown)


ruggedy:
1) regarding the code duplication, I am already working on figuring out shared functions and possibly moving them to shared/helpers. 
2) I have a couple ideas about supporting keyframes. hopefully I will have a working implementation before the weekend.

ruggedy:
are there any guidelines regarding project structure, procedures e.t.c

drcmda:
No, not so much, the code style is guided by prettier, but everything else is up to you.

My biggest question mark right now is how to handle multiple items b/c hooks can't be conditional/looped, which makes it harder for us. Controller can be looped but in itself it doesn't have keyframe capabilities.

ruggedy:
since we are talking about trails and keyframes. I will be exploring how we can make hooks for both of those as well. I will get back to you on my findings.

ruggedy:
you can check out the implementation of the trails hook [here.](https://codesandbox.io/s/xvo175z9ww). will be working on keyframes next and hopefully wrap up with transitions by the weekend.

drcmda:
It doesn't seem to move the circles in that box for now, but the code looks nice. Using multiple controllers is a good idea, that'll will take care of multiple items.

ruggedy:
yh was editing it thats why. you can have a look now. should work

drcmda:
Edited it, yeah i tried again and all is working. Looks perfect to me. How about we publish these hooks later on by putting them to /src/addons, we can worry later about code duplication. Like, when we have hooks for trail, keyframes and transitions, people could already test them in the wild. And in a second step trying to add them to /src by re-using as much code as we can.

ruggedy:
yeah that sounds like a great idea. would you prefer publishing them as soon as they are ready, or wait till all hooks are ready and publish them all in one go?

drcmda:
I guess all in one go would be better. Along with really simple examples (i could port some, and write some docs for it). Do you use twitter much? We could message there.

PR #360, opened 2018-12-07
Fix react-native transform

### LINKED ISSUE #361: Can't use React Native transform with animated(View)
Hi,

I've been trying to animate `translateY` on a RN View but it isn't animating at all.  i've been scratching my head for a while as to why it wasn't animating. 

What I found is not so much an issue with the animation itself, but it seems that `animated(View)` removes the `transform` style property entirely on the view. 

So, with:

`const AnimatedView = animated(View)`

and this:

```
<AnimatedView style={{ 
        borderWidth:1, 
        transform:[ { translateY:50 } ] 
}}>
   <Text>Hello</Text>
</AnimatedView>
```
The resulting view has borderWidth applied correctly but the transform isn't.

This is also validated but the RN debugger that shows only the borderWidth style is applied but the transform has been stripped out.

It doesn't actually matter whether or not I wrap this in a `<Spring></Spring>`, the result is the same.

I'm using RN 0.57 (in Expo 31), React 16.5.0 and react-spring 6.1.10

Thanks

Elindorath:
The transform property in style didn't work in react-native.
Resolves #361

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -19,9 +19,9 @@
     "gzipped": 1990
   },
   "dist/web.js": {
-    "bundled": 78895,
-    "minified": 37496,
-    "gzipped": 12279,
+    "bundled": 78206,
+    "minified": 37165,
+    "gzipped": 12177,
     "treeshaked": {
       "rollup": {
         "code": 30895,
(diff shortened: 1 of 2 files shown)


doginthehat:
I confirm this fixes #361

PR #376, opened 2018-12-13
[chore] remove overwritten property

samlanning:
The type specification for the `keys` property of Trail was defined twice (lines `10` and `18`), and defined differently in each case. This PR removes what looks like the incorrect definition (which also didn't match its docstring, as `keys` can be a function that maps an item to its key).

This will also result in no behavioural change, as when multiple properties are defined with the same name in an object literal, the last one is used. So this deleted code was being ignored anyway.

I found this in the alerts for this project on LGTM.com, and there are [a bunch of remaining alerts](https://lgtm.com/projects/g/drcmda/react-spring/alerts) that would probably also be good to look at.

*(Full disclosure: I'm part of the team behind LGTM.com)*

### DIFF
--- a/src/Trail.js
+++ b/src/Trail.js
@@ -6,10 +6,6 @@ import { toArray } from './shared/helpers'
 
 export default class Trail extends React.PureComponent {
   static propTypes = {
-    /** Item keys (the same keys you'd hand over to react in a list). If you specify items, keys can be an accessor function (item => item.key) */
-    keys: PropTypes.arrayOf(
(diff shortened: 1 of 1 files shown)


drcmda:
LGTM is awesome, never heard of it before but that seems insanely useful.

drcmda:
@samlanning btw, does it refresh automatically or does it need a new version tag or something? i've fixed most of these issues (some were false flags): https://github.com/drcmda/react-spring/commit/85b15f634ea4d987e0dfa57b24fa8179a714ddc9 but it still lists them

PR #398, opened 2018-12-22
More TypeScript updates

Kovensky:
Addresses what I could from https://github.com/react-spring/react-spring/issues/383, and adds types for `useChain`.

### DIFF
--- a/types/hooks.d.ts
+++ b/types/hooks.d.ts
@@ -1,3 +1,4 @@
+import { CSSProperties } from 'react'
 import {
   SpringConfig,
   SpringBaseProps,
@@ -57,8 +58,8 @@ export interface UseSpringBaseProps extends HooksBaseProps {
   config?: SpringBaseProps['config']
 }
 
-export type UseSpringProps<DS extends UseSpringBaseProps> = Merge<
-  DS,
+export type UseSpringProps<DS extends object> = Merge<
+  DS & UseSpringBaseProps,
   {
     from?: InferFrom<DS>
     /**
@@ -69,6 +70,12 @@ export type UseSpringProps<DS extends UseSpringBaseProps> = Merge<
 >
 
 // there's a third value in the tuple but it's not public API (?)
+export function useSpring<DS extends CSSProperties>(
+  values: UseSpringProps<DS & CSSProperties>
+): ForwardedProps<DS>
+export function useSpring<DS extends CSSProperties>(
+  getProps: () => UseSpringProps<DS & CSSProperties>
+): [ForwardedProps<DS>, SetUpdateFn<DS>]
 export function useSpring<DS extends object>(
   getProps: () => UseSpringProps<DS>
 ): [ForwardedProps<DS>, SetUpdateFn<DS>]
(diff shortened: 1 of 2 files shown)


types/hooks.d.ts
```diff
@@ -43,18 +48,34 @@ export type SetUpdateFn<DS extends object> = (ds: DS) => void
...
 }
 
 export interface UseSpringBaseProps extends HooksBaseProps {
   config?: SpringBaseProps['config']
 }
 
-export type UseSpringProps<DS extends UseSpringBaseProps> = Merge<
-  DS,
+export type UseSpringProps<DS extends object> = Merge<
+  DS & UseSpringBaseProps,
```
Kovensky: This change fixes autocomplete so it shows `UseSpringBaseProps`' properties on autocomplete

types/hooks.d.ts
```diff
@@ -65,6 +86,12 @@ export type UseSpringProps<DS extends UseSpringBaseProps> = Merge<
 >
 
 // there's a third value in the tuple but it's not public API (?)
+export function useSpring<DS extends CSSProperties>(
+  values: UseSpringProps<DS & CSSProperties>
```
Kovensky: This trick lets CSSProperties show up on autocomplete by default (as long as no property that you add is incompatible with CSSProperties itself; then it falls back to the other overloads)

types/universal.d.ts
```diff
@@ -64,7 +64,7 @@ export interface SpringBaseProps {
   onStart?(): void
 }
 
-export interface SpringProps<DS extends object = {}> {
+export interface SpringProps<DS extends object = {}> extends SpringBaseProps {
```
Kovensky: Added this to address https://github.com/react-spring/react-spring/issues/386

types/universal.d.ts
```diff
@@ -87,11 +87,6 @@ export interface SpringProps<DS extends object = {}> {
    * Takes a function that receives interpolated styles
    */
   children?: SpringRendererFunc<DS>
-  /**
-   * Prevents animation if true, you can also pass individual keys
-   * @default false
-   */
-  immediate?: boolean | string[] | ((key: string) => boolean)
```
Kovensky: Did `immediate` ever actually accept `string[]`? I can't find it in the code (in fact if it's an array it's just treated as always `true` in the `Controller`). This made it incompatible with `SpringBaseProps` so I had to remove the `string[]`.

drcmda: @Kovensky  No, immediate can be a boolean or a function returning boolean. This could be a left over.

Kovensky:
As I try to actually make use of this in a real project I noticed the type definition for `useTransition` is almost completely wrong 😓

I also wonder if it's possible to attach `.interpolate` to the output types. Might be again running into the "too generic for typescript" issue.

drcmda:
transitions return Array<{ item, key, props, state }>, wheresare state can be enter|leave|update and props is an object that contains animated values (which can interpolate).

Kovensky:
Yeah, the output type is (almost) correct, but also very wrong because the input type is wrong.

First, it assumes that, like `useSpring`, the `from` is derived from the input props but that seems to not be the case with `useTransition`. Second, it returns the input object itself (minus some filtered props) instead of the union of the types of `from` / `enter` / `update` / `leave`.

drcmda:
This we'll change soon, i guess it should be useTransition(input, keys, config), just like useTrail(items, config). Can we anticipate this change already and exclude items, or will TS make it impossible to use?

Kovensky:
I think it has to be done at the same time, because `items` is currently required to be able to infer a generic argument.

(Somewhat related: `initial` is not working correctly, `state.current.first` is still `false` the first time `transitions` is calculated so `initial` is only set on _not initial_ enters; the `initial: null` behavior is also unsupported)

cdebotton:
@Kovensky I apologize if you've already considered this, but I had fairly primitive types I was using for just `useSpring` that were working with initial release of `react-spring` with hooks. Perhaps these might be helpful?

https://github.com/cdebotton/cmyk/blob/master/app/types/react-spring.d.ts

cdebotton:
I've actually opened up a PR that adds the typings for chains.

https://github.com/react-spring/react-spring/pull/400

PR #400, opened 2018-12-22
Allow for interpolation chaining with hooks and TypeScript

cdebotton:
This PR adds an Opaque type that wraps `ForwardedRef` in the `react-spring` `hooks.d.ts` file. This allows for a type to be returned whose properties can be interpreted as valid CSS Properties as well as a type which can be interpolated.

eg:

```tsx
import { useSpring, animated } from 'react-spring/hooks';

interface Props {
  open?: boolean;
}

function App({ open = false }) {
  const { xy } = useSpring({
    xy: open ? [0, 0] : [1, 1],
  });

  // [number, number] => string
  const interpolateTransform = ([x, y]: [number, number]) => {
    return `translate3d(${x * 10}, ${y * 10}, 0px) scale(${x})`;
  };

  // [number, number] => number
  const interpolateBackgroundColor = ([x]: [number, number]) =>{
    return x;
  };

  return (
    <animated.div
      style={{
        backgroundColor: xy                               // [number, number]
          .interpolate(interpolateBackgroundColor)        // number
          .interpolate({
            range: [0, 1],
            output: ['#fff', '#000']
          }),                                             // string
        transform: xy.interpolate(interpolateTransform),  //string
      }}
    />
  );
}
```

*Additionally*, I've updated `SetUpdateFn` to exclude requiring the reserved words defined in `ExcludedProps`.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -19,83 +19,83 @@
     "gzipped": 1990
   },
   "dist/web.js": {
-    "bundled": 75825,
-    "minified": 36044,
-    "gzipped": 11686,
+    "bundled": 76049,
+    "minified": 36091,
+    "gzipped": 11709,
     "treeshaked": {
       "rollup": {
-        "code": 30418,
+        "code": 30465,
         "import_statements": 287
       },
       "webpack": {
-        "code": 31818
+        "code": 31865
       }
     }
   },
   "dist/web.umd.js": {
-    "bundled": 81489,
-    "minified": 32103,
-    "gzipped": 11503
+    "bundled": 81727,
+    "minified": 32150,
+    "gzipped": 11528
   },
   "dist/native.js": {
-    "bundled": 71944,
-    "minified": 33557,
-    "gzipped": 10299,
+    "bundled": 72168,
+    "minified": 33604,
+    "gzipped": 10326,
     "treeshaked": {
       "rollup": {
-        "code": 26924,
+        "code": 26971,
         "import_statements": 298
       },
       "webpack": {
-        "code": 28273
+        "code": 28320
       }
     }
   },
   "dist/universal.js": {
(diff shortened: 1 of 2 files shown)


chandana111:
A simple implementation of chain in VanillaJS
Let’s start with a quick implementation of something like chain in VanillaJS:
// This is an implementation of chain() in VanillaJS.
// It supports map(), reduce(), mapValues() and has an auto-closing sum() method.
class WrappedValue {
  constructor(v) {
    this.v = v;
  }

  filter(fn) {
    this.v = this.v.filter(fn);
    return this;
  }
  reduce(fn, x) {
    this.v = this.v.reduce(fn, x);
    return this;
  }
  map(fn) {
    if (typeof(fn) === 'string') {
      this.v = this.v.map(v => v[fn]);
    } else {
      this.v = this.v.map(fn);
    }
    return this;
  }
  mapValues(fn) {
    for (const k in this.v) { this.v[k] = fn(this.v[k], k); }
    return this;
  }
  sum(fn) {
    return this.reduce((a, b) => a + b, 0).value();  // auto-close
  }
  value() {
    return this.v;
  }
}

function chain(value) {
  return new WrappedValue(value);
}

// (This isn’t at all how Lodash’s implicit chain is implemented.
// It uses lazy evaluation to get some nice optimizations.)

[angularjs training in Marathahalli
](url)
[ angularjs training institutes in Marathahalli](url)

[ best angularjs training in Marathahalli](url)

Kovensky:
Actually, one thing I never tested is, what happens if you try to animate an object? Say, animate `{ x2: number; y2: number; style: { stroke: string } }`; should the result of that work when spread onto an `<animated.rect/>`?

Depending on how it's implemented it could allow us to not special-case `style` and instead just support arbitrary object properties.

Issue #401, opened 2018-12-23
Uncaught TypeError: entry.interpolation.calc is not a function when I try to use transition with left prop and mix numbers and strings


capaj:
I tried animating left prop in the transition page example: https://codesandbox.io/s/nk219j7pl4
 instead of `translate3d` and I got this typeError:

```
Uncaught TypeError: entry.interpolation.calc is not a function
    at eval (nk219j7pl4.codesandbox.io/node_modules/react-spring/web.cjs.js:1388)
    at Array.reduce (<anonymous>)
    at Controller.update (nk219j7pl4.codesandbox.io/node_modules/react-spring/web.cjs.js:1368)
    at Spring.render (nk219j7pl4.codesandbox.io/node_modules/react-spring/web.cjs.js:1800)
    at finishClassComponent (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:14301)
    at updateClassComponent (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:14264)
    at beginWork (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:15082)
    at performUnitOfWork (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:17820)
    at workLoop (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:17860)
    at HTMLUnknownElement.callCallback (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:199)
    at invokeGuardedCallback (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:256)
    at replayUnitOfWork (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:17107)
    at renderRoot (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:17979)
    at performWorkOnRoot (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:18837)
    at performWork (nk219j7pl4.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:18749)
[...]

PR #402, opened 2018-12-23
adds a TypeError to help new users avoid mixing strings and numbers

capaj:
as suggested in #401

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,6 +11,7 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    "babel-plugin-dev-expression",
     ["@babel/proposal-class-properties", { "loose": true }],
     ["@babel/plugin-proposal-object-rest-spread", { "loose": true }],
     ["transform-react-remove-prop-types", { "removeImport": true }]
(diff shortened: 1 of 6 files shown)


drcmda:
Nice idea, how about using "invariant" like the original animated (https://github.com/animatedjs/animated/blob/master/src/Interpolation.js). I think invariant will only apply to dev env and get removed in production.

capaj:
Will amend it

drcmda:
Maybe even https://www.npmjs.com/package/tiny-invariant , that way we don't inline the sprintf logic that invariant has.

capaj:
@drcmda I am not sure if the build itself doesn't need more tweaking-I just added the `babel-plugin-dev-expression` plugin to babelrc

drcmda:
I'll keep this up for a little while longer, i am thinking - maybe it's a good idea to add more of these warning, especially for interpolation

aleclarson:
Closing this since v9 now warns when an interpolation fails:
https://github.com/react-spring/react-spring/blob/512f1e849d6d94729f0fce929e531e1a3249a318/packages/core/src/Controller.ts#L590-L595

PR #440, opened 2019-01-12
Emphasize docs in README

joshwcomeau:
I use React Spring quite a bit these days, which means I'm often looking to check the docs for specific APIs. And for some reason, I always have a really tough time actually finding them! 

The docs site doesn't show up at all when you google "React spring docs":

docs

After clicking "Documentation" in that first result, I wind up in this repo, except it's a bunch of markdown files. So I navigate up to the main README, and scroll around looking for the link to the docs. This takes me several tries, since my quick-scanning brain totally misses this link:

easytomiss

Here's what it looks like in this PR:

screen shot 2019-01-12 at 10 03 32 am

I know doing READMEs is tricky because you're trying to match several usecases at once, from the beginner who doesn't know what this even is, to the returning users looking for the API reference docs. But I think the docs/examples are universally a good thing to highlight, especially when they're as nice as the ones you built!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -40,7 +40,7 @@ For a more detailed explanation read [Why React needed yet another animation lib
 
 # Documentation and Examples
 
-Full documentation and examples here: [react-spring.surge.sh](https://react-spring.github.io)
+Full documentation and examples here: **[react-spring.surge.sh](https://react-spring.github.io)**
 
 ---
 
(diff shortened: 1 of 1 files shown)


drcmda:
Could you point it to the surge.sh url? It's linking the github page, but that one doesn't support SPA/routing.

joshwcomeau:
Thanks! I've already used the new link twice =D

PR #450, opened 2019-01-16
I believe the Keyframe needs to pass the delay

JarrodMFlesch:
Does this change make sense? I had to add it on my local copy to get anything to work correctly. I just want to make sure others don't struggle on something so small because a word is missing.


drcmda:
i think the api signature is wrong, or should be. delay is a prop. what caused a bit of a problem (and still does in hooks) is that props should be accumulative. you set delay: 600 once, it should merge that internally and even if you omit it later on, it's going to be 600ms unless you explicitly cancel it out.

aleclarson:
Closing this since the `Keyframes` component is about to be a legacy API.

Issue #447, opened 2019-01-16
useImperativeMethods → useImperativeHandle


phaistonian:
It seems the latest `next` react version replaced `useImperativeMethods` with `useImperativeHandle `.

This is causing `useTransitions` (at least) to break.

j8seangel:
Yes, I was having the same issue and this is the commit included in the latest alpha.1 version.
https://github.com/facebook/react/commit/b4ad8e947150a1a0b486a388e2d4762d3eee51ee

Fixed using react `16.8.0-alpha.1` for now

drcmda:
could you submit a quick PR? i am in the middle of a refactor and wouldn't want to mess with the local state.

jnsdls:
also breaks `useSpring` and `useSprings`

@drcmda I'll have a PR for you in a minute here

jnsdls:
It’s proving more complicated than expected. (Not as simple as just switching out the function names and calling it a day.)

*so disregard the “in a minute” part of my previous comment*

jnsdls:
@drcmda I have that PR now: https://github.com/react-spring/react-spring/pull/451

Jessidhia:
`const { useImperativeHandle = React.useImperativeMethods } = React` _should_ be sufficient 🤔

BenoitZugmeyer:
For those stumbling on this issue, here is a quick solution before a proper react-spring fix is landing.

In your App entry point, just add:
```js
import React from 'react'
React.useImperativeMethods = React.useImperativeHandle
```

JimmyRowland:
`const { useImperativeHandle = React.useImperativeMethods } = React` 
is not working for me.
8.0.0-beta.3 works fine, but I am using parallax for a different component and parallax no longer exists under 'react-spring/addons. Is parallax bugged because of the new react update?

JimmyRowland:
Thanks for the update, but parallax is still not working.
Here is the error message:
https://codesandbox.io/s/z32qz2n69p

drcmda:
some export missing, will fix it

phaistonian:
@drcmda Will this change in export land in `b5` or it applied already?

drcmda:
@JimmyRowland fixed the export, it looks good now: https://codesandbox.io/s/48r42p59o7 just keep in mind, you can't mix any exports, like using "animated" from react-spring in react-spring/renderprops-xxxx. they are separate libraries.

@phaistonian beta.5, correct

drcmda:
[...]

PR #451, opened 2019-01-17
Hooks: switch useImperativeMethods -> useImperativeHandle (react 16.8.0-alpha.1)

jnsdls:
I assumed this would be as easy as switching some import names...
_It was not_

This PR does the following:
1) switch useImperativeMethods -> useImperativeHandle (should close: #447 )
2) reverts part of https://github.com/react-spring/react-spring/commit/d18922b58ebf6ecdc9c19f3e1ec9afff0c586428 (more about why below)
3) upgrades the dependency (for hooks) to `react@16.8.0-alpha.1`
4) fixes the gestures-pager example (which would "crash" the page on unmount) fixes: https://github.com/react-spring/react-spring/issues/429

Why 2)? Why the revert?

When I was working on 1) I used the docz hooks examples page as kitchen sink kind of test scenario, if you will. I noticed that it would crash, did a git bisect to find the commit that introduced the breakage and found https://github.com/react-spring/react-spring/commit/d18922b58ebf6ecdc9c19f3e1ec9afff0c586428 .
By removing the changes to the KeyframeController I was able to fix the crash. @ruggedy I'd love your feedback on this. (I tried to fix the crash with keeping the change but after a while I gave up. (Especially because after reverting everything **I** tested worked.)

Is this tested?
A bit, definitely not exhaustively.

I'm happy to work on this more (over the weekend), especially if I can get some pointers of how to fix the problem RE the `KeyframeController` without reverting that change.


drcmda:
hey @jnsdls thanks for all the work! im really sorry i didn't know you were going so deep - i am currently rewriting the lib (see the hooks-8.x branch) - keyframescontroller doesn't exist any longer in there and props handling is all central to the controller.

jnsdls:
@drcmda happy to hear it, how close are u to a (alpha?) release of that. I honestly just took the opportunity to poke around and learn more about some internals anyways.

Is there something I can track for development, or a list of stuff you may want help with?

Let me know if you want me to send a separate PR to just fix the `gestures-pager` that causes crashes on the /hooks route in the docs in the meantime.

brunolemos:
When the rename gets merged please don't make it a breaking change, make it in a way both versions of react would work, otherwise it will be hard to update the dependencies that use hooks, not all are updated to handle this

jnsdls:
@drcmda I'll close this PR since it's superseded by your work on the hooks-8.x branch. I think one change that would be good to make in the meantime is to add info to the hooks docs page that it requires react@16.8.0-alpha.0 (and breaks with react@16.8.0-alpha.0 and up). I'll send a PR for that later.

drcmda:
@jnsdls if you want to help, of course! and it's great you've been digging into it already - there's definitively a lot of things that need some work. 

@brunolemos you mean uIM -> iIH? or hooks in general? hooks will be a new export, i don't plan to mix them with classes so nothing will break. i won't mess around with classes when hooks are out, personally i mean, i am already doing most of my personal projects on hooks and i can't imagine i'll move back. This will have affect react-spring as well b/c i think hooks really are a better way to handle animation - i will most likely focus on it.

brunolemos:
@drcmda great to hear because I also use only hooks nowadays! 

E.g. https://github.com/devhubapp/devhub (will deploy the version with react-spring soon, made the app feel much nicer to use)

brunolemos:
@drcmda what I mean about the rename is that there will be people and libs out there using react alpha 16.7 and some others using 16.8, if you rename the uIM please make in a way that works on both versions.

For example, something like this should work: `const uIH= React.useImperativeHandle || React.useImperativeMethods`

PR #453, opened 2019-01-17
fix gestures-pager demo

jnsdls:
change prevents a crash on unmounting of this component


jnsdls:
@drcmda pulled this out of #451 because it's an easy change that fixes the hooks docz page navigation

PR #455, opened 2019-01-19
Create native-hooks.d.ts

brunolemos:
Fix missing type definition for `native-hooks` target.

### DIFF
--- a/types/native-hooks.d.ts
+++ b/types/native-hooks.d.ts
@@ -1,0 +1,1 @@
+export * from './hooks'


drcmda:
hey @brunolemos since you use the lib professionally, would you want access to the repo with maintenance/team status? it would allow you to make such changes easier and i could use some help since i'm preparing 8.x so i can't concentrate much on the 7.x build.

brunolemos:
@drcmda it would be cool to be added to the repository :) but I don't think much would change in practice. I can't be a real maintainer because I'm focusing on DevHub, so I'd contribute only to the issues that I face directly, and I'd probably make PRs even if I had push access to master.

PR #457, opened 2019-01-20
[react-native] Support nested array styles

### LINKED ISSUE #439: [react-native] Support style as array

### LINKED ISSUE #456: native-hooks not working
I was able to use `react-spring/hooks` on react-native, with some limitations, of course.

But when I try to use `react-spring/native-hooks` it doesn't work at all :(

#### 1. Wrong import 

Made #454

#### 2. Missing type definitions

Made #455 

#### 3. Frozen object?
```
Error: You attempted to set the key `props` with the value `{"config":{"duration":100},"native":true,"backgroundColor":"rgba(77,83,96,0)","interpolateTo":true,"autoStart":true}` on an object that is meant to be immutable and has been frozen.
```

```
You attempted to set the key `done` with the value `true` on an object that is meant to be immutable and has been frozen.
```

Code is something like:

```tsx
const [springAnimatedStyles, setSpringAnimatedStyles] = useSpring(() => {
    const { isHovered, theme } = cacheRef.current

    return {
      config: { duration: 100 },
      native: false,
      backgroundColor:
        isHovered
          ? theme.backgroundColorLess16
          : theme.backgroundColorLess08,
    }
  })
```

#### 4. When passing an array to style, it's converting it to an object with number keys

```
Failed prop type: Invalid props.style key `0` supplied to `Text`.
Bad object: {
  "0": {
    "fontWeight": "600",
    "fontSize": 15,
    "lineHeight": 20,
    "textAlign": "left"
  },
  "1": {}
}
```

brunolemos:
Fix #439
Fix #456

This also fixes `react-native-web` support.

> This is an alternative PR to #458. Check pros and cons below. Pick one.

### Demo

- https://codesandbox.io/s/jv55n5yyow
- https://snack.expo.io/@brunolemos/react-spring-styles-array

### Description

React Native not only supports `<View style=[{}, {}] />`, but also nested arrays like `<View style={[{}, [{}, [{}]]]} />`, this situation happens quite easily. 

`StyleSheet.flatten` exists to handle this, it merges all arrays into a single object.

The problem with this approach it that it loses the optimization that `react-native-web` does. When doing `StyleSheet.create({ container: {} )`, the result will be `{ container: 1 }` on `react-native-web`, an int ID. _(`react-native` used to have this, but they removed it temporarily, may be added again in the future)._

I made another PR that preserves the style ids: #458 

### Pros compared to #458 

- Cleaner implementation, less likely to cause bugs in the future.

### Cons

- Loses the style id optimization. (note: style ids were already not supported by react-spring, now they will be transformed to style objects)


brunolemos:
This implementation is cleaner but loses optimization, so I'll leave only #458 open. You can choose any of the two that you prefer.

drcmda:
Is it done? Should i merge and release a patch? Thanks btw!!

brunolemos:
Yes

brunolemos:
Thanks! Just [tweeted](https://twitter.com/brunolemos/status/1087456685080358918?s=20) about it.

kations:
It seems that this only works in render props? Would be great to have support in hooks too.

drcmda:
The code in this PR is merged, Stylesheet.flatten is being used, no matter if hooks or renderprops.

kations:
It don't seems to work with:

const AnimatedView = animated(View);

const { rotate, opacity } = useSpring({
    from: { opacity: 0, rotate: 0 },
    to: async next => {
      while (true) {
        await next({ opacity: 1, rotate: 360 });
        await next({ opacity: 0, rotate: 0 });
      }
    }
  });

and 

<AnimatedView
      style={{
        width: size,
        height: size,
        opacity: opacity,
        transform: [{ rotate: rotate + "deg" }]
      }}
    >

only the opacity is animating. Version: "react-spring": "^8.0.7",

drcmda:
shouldn't this be rotateX or Y?

drcmda:
@kations rotate is not a value, doing `rotate + "deg"` is like adding a string to a class. you can either do rotate: '0deg' -> '360deg' above in the spring, or interpolate inside the view:

 transform: [{ rotate: rotate.interpolate(r => `${r}deg`) }]

see: https://www.react-spring.io/docs/hooks/basics

kations:
@drcmda thanks for the feedback, is it possible that this only works in react-native? My use case is a universal component library based on react-native-web. While it works in react-native it didn't with the same component renders in the browser.

kations:
@drcmda any info how to solve the use case with react-native-web?

kations:
found a solution by interpolation and returning the whole transform array

drcmda:
@kations i haven't worked with native web before, i assumed it works in the same way as RN. But it seems it doesn't. in that case a new target could make sense. if you like, open src/targets/native and take a look. you could submit another PR for native-web users.

PR #458, opened 2019-01-20
[react-native] Support nested array styles

### LINKED ISSUE #439: [react-native] Support style as array

### LINKED ISSUE #456: native-hooks not working

brunolemos:
Fix #439
Fix #456

This also fixes `react-native-web` support.

> This is an alternative PR to #457. Check pros and cons below. Pick one.

### Demo

- https://codesandbox.io/s/jv55n5yyow
- https://snack.expo.io/@brunolemos/react-spring-styles-array

### Description

React Native not only supports `<View style=[{}, {}] />`, but also nested arrays like `<View style={[{}, [{}, [{}]]]} />`, this situation happens quite easily. 

`StyleSheet.flatten` exists to handle this, it merges all arrays into a single object. I used it on #457 as an alternative to this PR.

The problem with that approach it that it loses the optimization that `react-native-web` does. When doing `StyleSheet.create({ container: {} )`, the result will be `{ container: 1 }` on `react-native-web`, an int ID. _(`react-native` used to have this, but they removed it temporarily, may be added again in the future)._

### Pros compared to #457

- This PR preserves the stylesheet id optimization.

### Cons

- More verbose implementation. `this.payload` is now a mix of numbers and `AnimatedStyle` instances. 

_Not sure if this would break things in the future. This number check could be moved to another place as well, you know better where it should be._

### DIFF
--- a/src/shared/helpers.js
+++ b/src/shared/helpers.js
@@ -104,3 +104,15 @@ export function handleRef(ref, forward) {
   }
   return ref
 }
+
+export function flattenDeep(array) {
+  const items = toArray(array)
+
+  return items.reduce(
+    (result, item) =>
+      Array.isArray(item)
+        ? result.concat(flattenDeep(item))
+        : result.concat(item),
+    []
+  )
+}
--- a/src/targets/native/AnimatedStyles.js
+++ b/src/targets/native/AnimatedStyles.js
@@ -1,25 +1,36 @@
 import AnimatedWithChildren from '../../animated/AnimatedWithChildren'
 import AnimatedStyle from '../../animated/AnimatedStyle'
+import { flattenDeep } from '../../shared/helpers'
 
 export default class AnimatedStyles extends AnimatedWithChildren {
   constructor(styles) {
     super()
-    this.payload = styles.map(style => new AnimatedStyle(style))
+    this.payload = flattenDeep(styles).map(style =>
+      typeof style === 'number' ? style : new AnimatedStyle(style)
+    )
   }
 
   getValue() {
-    return this.payload.map(style => style.getValue())
+    return this.payload.map(style =>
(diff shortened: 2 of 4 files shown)


brunolemos:
I just noticed a difference in behavior when using this PR. Something like this:

```
style={[{ color: '#FF0000' }, { color: springAnimatedStyles.color }]}
```

This was supposed to show the spring color but was not.

We could fix this here but would make the code even more verbose.
We better use the other implementation.

Issue #461, opened 2019-01-21
Transition mixes up element order when multiple items are removed simultaneously


mart-jansink:
When you remove multiple items simultaneously, the reordering logic at https://github.com/react-spring/react-spring/blob/master/src/Transition.js#L163 fails to get it correct. This results in React creating new elements which in turn resets e.g. the scroll offset of a div (which was the symptom that lead me to this issue).

I've reproduced it at https://codesandbox.io/s/10ol3v9oj, but you'll need to add a breakpoint at line 2348 of `https://10ol3v9oj.codesandbox.io/node_modules/react-spring/dist/web.cjs.js` like it is shown below:

![screen shot 2019-01-21 at 10 20 39](https://user-images.githubusercontent.com/4381258/51464604-3e43a100-1d66-11e9-9d46-72383bd82c67.png)

At the step that both item 1 and 2 are removed at once the following happens:
- first item 1 is removed and deleted correctly, after which item 0 and item 2 are still rendered in the correct order
- item 2 is removed but during its removal first item 2 and then item 0 are rendered, as you can see in the elements tab of the console and as indicated by the red 0

As far as I can tell this is because the `onRest` callback is first called for item 1, which results in a rerender before the `onRest` callback is called for item 2. This then results in another rerender but the reordering logic can't find the correct position for item 2 because item 1 is already gone.

Ideally, both items would be deleted at once, preventing the extra rerender all together. Somehow the order in which the `onRest` callbacks fire would also solve the issue, but I couldn't find a good way to do so in the code. For me, the actual order of the rendered elements doesn't really matter (they're absolutely positioned) as long as it doesn't change, so I could solve it by calling `.reverse()` on the array of items before passing them as a prop.

drcmda:
The demo uses an old version of react-spring, is it the same with the lastest version as well (7.x)?

mart-jansink:
Yes it is. My apologies, I forked this sandbox from one of the demos. It's now updated and I still see the issue when looking at the elements tab of the console:

![screen shot 2019-01-22 at 11 17 04](https://user-images.githubusercontent.com/4381258/51528595-491a3680-1e37-11e9-9ccc-ac02f67c407b.png)

aocenas:
Have the same issue. When 2 or more items which are next to each other are removed this code https://github.com/react-spring/react-spring/blob/master/src/useTransition.js#L252 will put all but one of them at the beginning of the array because `left` item is already deleted.

[...]

PR #462, opened 2019-01-21
Batch deletion of removed items

mart-jansink:
Batch the deletion of removed items after they transitioned out. See #461.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import Spring from './Spring'
 import Keyframes from './Keyframes'
+import { requestFrame } from './animated/Globals'
 import { callProp, toArray, interpolateTo } from './shared/helpers'
 
 let guid = 0
@@ -74,6 +75,7 @@ export default class Transition extends React.PureComponent {
       deleted: [],
       prevProps,
     }
+    this.batch = null
   }
 
   static getDerivedStateFromProps(props, { first, prevProps, ...state }) {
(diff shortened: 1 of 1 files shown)


drcmda:
could you checkout the hooks.8.x branch and see if this is still an issue? it's going to get all refactored soon and in the new branch it removes lazily by default already

grrowl:
I'm still seeing this issue on `react-spring@8.0.8` using the `useTransition` hook. Would love to see this merged :)

drcmda:
@grrowl could you make a minimal codesandbox to show the effect?

grrowl:
Sure thing: this demo exhibits the behavior using hooks and the latest react-spring: https://codesandbox.io/s/my869ojmjj

drcmda:
@grrowl could you give me a super quick rundown how you expect it to behave?

grrowl:
Sorry, I didn't explain the problem exhibited by the demo.

When you click "Filter", we re-render with only the item "b". But notice, the item "b" moves into the space taken by "f", even though all the children are keyed and should maintain their position through transitions in and out. 

So, what we expect when changing the items array is rendering a list of (using punctuation to illustrate fading: 

* `[a] [b] [c] [d] [e] [f] [g]` => begin transition
* `.a. [b] .c. .d. .e. .f. .g.` => transition finishes
* `[b]`

The problem is the items are re-ordered when rendered through the "transition" phase as:

* `[a] [b] [c] [d] [e] [f] [g]` => begin transition
* `.a. .d. .e. .f. .g. [b] .c.` => transition finishes (`b` and `c` move to the end of the `transitions` array)
* `[b]`

In my example I included the `key` in item render since, especially when setting `unique: true`, elements' order should be maintained throughout renders.

drcmda:
yeah, i think batch delete wouldn't have an effect on order, and it actually does batch delete now in some situations. if you want to look at it, that would be amazing as my time's limited right now. it's here somewhere: https://github.com/react-spring/react-spring/blob/master/src/useTransition.js#L250-L259

here's how i debug stuff: https://github.com/react-spring/react-spring-examples/blob/master/README.md

PR #466, opened 2019-01-24
chore(docs): document Transition config by key

ostrgard:
I had to read the source to understand that conditionally setting config by key had not been removed. I think this clarifies it better.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -43,7 +43,7 @@ export default class Transition extends React.PureComponent {
     unique: PropTypes.bool,
     /** Useful in combination with "unique", when true it forces incoming items that already exist to restart instead of adapting to their current values */
     reset: PropTypes.bool,
-    /** Spring config, or for individual keys: fn((item,type) => config), where "type" can be either enter, leave or update */
+    /** Spring config, or for individual keys: fn((item,type) => key => config) or fn((item,type) => config), where "type" can be either enter, leave or update */
     config: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
     /** Calls back once a transition is about to wrap up */
     onDestroyed: PropTypes.func,


eoghanmccarthy:
@ostrgard Do you have a working example of setting the transition config by type (enter, leave)? Struggling to understand it?

ostrgard:
In `Transition` the `config` prop is a function that loops over `items`. You can either return a `config` or a function which the keys will be looped over. Maybe the docs could be even more explanatory than what I propose in this PR. Here's an example I hope you can follow:

```
...
items={[...]}
from={{width: 0, opacity: 0}}
to={{width: 20, opacity: 1}}
config={(item, type) => key => key === 'width' ? config.default : config.wobble}
...
```

Feel free to ask if it doesn't help you.

eoghanmccarthy:
@ostrgard Thanks, I'm trying to set the config particularly for the 'leave' event. Going by the docs I would guess that the transition config should be as below (I think), however, it doesn't seem to work. Any ideas to fix?

```
config={(item, type) => key =>
              key === 'leave'
                ? { tension: 20, friction: 12, clamp: true }
                : config.default}
```

ostrgard:
Ah, I see. `key` is for mapping over the keys your specify in `to`, `from` etc. That could be `width` or `opacity`, like in my previous example. `type` is the "event" or whatever you want to call it.

This is what you want to do:
```
config={(item, type) =>
              type === 'leave'
                ? { tension: 20, friction: 12, clamp: true }
                : config.default}
```

eoghanmccarthy:
@ostrgard Thanks, that works.

Just a general question, if you're swiping in a single modal from off-screen right to centre and back out, for example, is it better to use Spring? Is Transition mainly for when you have more than one element that you need to transition between such as photos in a gallery?

ostrgard:
> Just a general question, if you're swiping in a single modal from off-screen right to centre and back out, for example, is it better to use Spring? Is Transition mainly for when you have more than one element that you need to transition between such as photos in a gallery?

Whenever you want something to appear and later disappear, use `Transition`. Whenever you want something to have a continuous motion, use `Spring`. That's how I think about it. A modal is a great use case of a `Transition` as it follows component life cycle.

drcmda:
Good catch, thanks! Will update docs soon ...

drcmda:
Btw, with hooks this will become a little more logical, since spring itself can accept functions and array chains, the others merely inherit it. keyframes and transitions used to be special cases.

PR #471, opened 2019-01-26
Typo fix

ElForastero:


drcmda:
Thanks! In 8.x using hooks the flag won't be necessary any longer, if the async function resolves onRest will be called.

PR #496, opened 2019-02-06
Require `to` prop to correctly extend SpringProps

BrendanBerkley:
When upgrading today, I got this error:

```
ERROR in [at-loader] ./node_modules/react-spring/renderprops-universal.d.ts:234:11 
    TS2430: Interface 'TrailProps<TItem, DS>' incorrectly extends interface 'SpringProps<{}>'.
  Property 'to' is optional in type 'TrailProps<TItem, DS>' but required in type 'SpringProps<{}>'.
```

I opened a PR to correct this error, but for unrelated reasons I was unable to confirm the fix in the project where I'm using react-spring. So I tried running tests and wasn't able to pass `yarn test` or `yarn test:ts`. 

Here are the TS errors for 8.0.1: 

```bash
brendan ~/P/t/react-spring> yarn test:ts
yarn run v1.13.0
$ tsc --noEmit
types/native-hooks.d.ts:1:15 - error TS2307: Cannot find module './hooks'.

1 export * from './hooks'
                ~~~~~~~~~

types/renderprops-konva.d.ts:6:24 - error TS2307: Cannot find module 'react-konva'.

6 import * as konva from 'react-konva'
                         ~~~~~~~~~~~~~

types/renderprops-konva.d.ts:14:5 - error TS2538: Type 'symbol' cannot be used as an index type.

14   }[keyof typeof konva]
       ~~~~~~~~~~~~~~~~~~

types/renderprops-konva.d.ts:20:29 - error TS2344: Type 'Pick<any, string>[Tag]' does not satisfy the constraint 'ReactType<any>'.
  Type 'Pick<any, string>[Tag]' is not assignable to type '"mpath"'.

20       ComponentPropsWithRef<KonvaComponents[Tag]>
                               ~~~~~~~~~~~~~~~~~~~~

types/renderprops-universal.d.ts:234:11 - error TS2430: Interface 'TrailProps<TItem, DS>' incorrectly extends interface 'SpringProps<{}>'.
  Types of property 'children' are incompatible.
    Type '((item: TItem, index: number) => SpringRendererFunc<DS>) | undefined' is not assignable to type 'SpringRendererFunc<{}> | undefined'.
      Type '(item: TItem, index: number) => SpringRendererFunc<DS>' is not assignable to type 'SpringRendererFunc<{}>'.

234 interface TrailProps<TItem, DS extends object = {}> extends SpringProps {
              ~~~~~~~~~~


Found 5 errors.

```

After my one-character fix, the last error is replaced with this:

```bash
types/renderprops-universal.d.ts:234:11 - error TS2430: Interface 'TrailProps<TItem, DS>' incorrectly extends interface 'SpringProps<{}>'.
  Types of property 'children' are incompatible.
    Type '((item: TItem, index: number) => SpringRendererFunc<DS>) | undefined' is not assignable to type 'SpringRendererFunc<{}> | undefined'.
      Type '(item: TItem, index: number) => SpringRendererFunc<DS>' is not assignable to type 'SpringRendererFunc<{}>'.

234 interface TrailProps<TItem, DS extends object = {}> extends SpringProps {
              ~~~~~~~~~~

```

Wasn't sure how to fix that. I did notice that [this commit](https://github.com/react-spring/react-spring/commit/6ce0934928febd5f3bef866a50897bbbfffd1294) seems to be the what broke the types for this interface.

Hope I helped a bit!

### DIFF
--- a/types/renderprops-universal.d.ts
+++ b/types/renderprops-universal.d.ts
@@ -239,7 +239,7 @@ interface TrailProps<TItem, DS extends object = {}> extends SpringProps {
   /**
    * Animates to ...
    */
-  to?: DS
+  to: DS
   /**
    * An array of items to be displayed, use this if you need access to the actual items when distributing values as functions
    */


drcmda:
so, is this good to go? i wish we'd have a dedicated TS person around here 😇

BrendanBerkley:
I can assert with reasonable confidence that my one-character merge won't make anything worse 😎 

It won't resolve all the TS errors, though. I'm willing to help! But I'll need some direction.

The first error that I pasted above is related to types/native-hooks.d.ts. I poked around a bit. Guessing that types/native-hooks.d.ts can be deleted? Looks like the former contents of that hooks file was moved into web.d.ts, making this irrelevant?

Errors 2, 3,  and 4 are in renderprops-konva.d.ts. My guess was that react-konva needed to be added as a dependency. Is this correct? Once `import * as konva from 'react-konva'` has react-konva to resolve to, these three go away. (A new one arises, but it might be more of a react-konva problem than a react-spring problem.)

Error 5 is what my one-character PR attempted to fix, and my other code snippet is the new error that happens once I did so. There's a type mismatch with `children`, and I don't know how to fix that. The easiest way to fix everything is to undo my current commit and change `extends SpringProps` to `extends SpringBaseProps`, but as I mentioned in my first comment, that'd be undoing a specific change. So I don't know how to proceed here.

Regarding hooks and konva, I have some local changes ready that I can commit if you think they're the right call. Let me know!

PR #511, opened 2019-02-07
Fix useTransition type definition

### LINKED ISSUE #510: Typescript error for useTransition
I'm using React with Typescript and getting typing errors for useTransition:
```
const items = [{id: 1}, {id: 2}];
const transitions = useTransition(items, item => item.id, {
  from: { opacity: 0 },
  enter: { opacity: 1 },
  leave: { opacity: 0 },
})
```
Resulting in "Expected 1 arguments, but got 3."
I'm not sufficient enough in writing types, would be nice if someone can add them.

brunolemos:
Fix #510

Complements #507

### DIFF
--- a/types/web.d.ts
+++ b/types/web.d.ts
@@ -163,13 +163,6 @@ export interface UseTransitionProps<TItem, DS extends object>
    * @default config.default
    */
   config?: SpringConfig | ((item: TItem) => SpringConfig)
-  /**
-   * The same keys you would normally hand over to React in a list. Keys can be specified as a key-accessor function, an array of keys, or a single value
-   */
-  keys?:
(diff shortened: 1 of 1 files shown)


zhammer:
it seems like this doesnt support the case where `items` is a boolean to toggle visibility:

https://github.com/react-spring/react-spring/issues/495#issuecomment-461192304

zhammer:
example from @drcmda 

```js
const [state, toggle] = useState(true)
const transitions = useTransition(state, null, { ... })

transitions.map(({ item, key, props }) => item && <animated.div key={key} style={props}> ...
```

jigsawye:
@zhammer I have a pull request in #531 to fixed it

PR #523, opened 2019-02-09
Update basics.mdx

christiannwamba:


drcmda:
The shortcut is valid though, any reason you want it to be the long-form?

christiannwamba:
Yeah. Just consistency with the demo. Nothing else.

I got confused when I saw that the demo was different from the code example in the doc until I went back to re-read these lines: https://github.com/react-spring/react-spring/blob/master/documentation/basics.mdx#L220-L223

You can ignore it though.

PR #537, opened 2019-02-12
Fix scrollTop type in animated

jacobrask:


drcmda:
good catch, this is serious bug : -S

PR #540, opened 2019-02-14
Fix useSpring result type

vlazh:


drcmda:
i dont understand ts good enough but sure this is how it's supposed to be? useSpring returns an objects with name:value pairs, the values are animatedValues.

vlazh:
Yes, values are AnimatedValues. And the PR is made for support it for type checking and autocomplete.

drcmda:
just asking to make sure :) thanks for the PR!

PR #543, opened 2019-02-14
Make useSpring support for css and not css properties as well without…

vlazh:
… type checking errors


drcmda:
will it still be able to do

    useSpring({ iCanAnimateAnything: 123, opacity: 1 })

?

vlazh:
Yes, it will.
image

vlazh:
And you can still use autocomplete for css properties:
image

PR #557, opened 2019-02-19
Bad package name in package.json

morgs32:
Collateral from a sloppy PR


drcmda:
no worries, i fixed it locally. just testing it out some more and will update soon :)

PR #558, opened 2019-02-19
Convert Controller, Frameloop and target files to TypeScript

jacobrask:
Controller types are pretty low quality still, but I need to get the entire code base to TypeScript to start experimenting with https://github.com/Swatinem/rollup-plugin-dts and other solutions to get output a type bundle.

There is a breaking change here in that `animated.marquee` is removed, because it is not recognized as a JSX element and thus not compatible with `ReactType`. We can either

1) just remove it, it's a deprecated element which is probably why it's not in the JSX types
2) try adding it with a PR to the React JSX types
3) maintain our own types which is basically `JSXElements` + `marquee`...


drcmda:
This pull request **fixes 10 alerts** when merging c651c2b353635f2610d212693254f59f03aa3a3f into 74b0984a89bdcf03fb4ea529a11e3d01396fd03c - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-d5c10ba2ebc785add8a70131f3e01eb8d46ae576)

**fixed alerts:**

* 10 for Unused variable, import, function or class

---

drcmda:
i think removing marquee is fine. is this good to go?

PR #561, opened 2019-02-20
share code between hooks and classes

drcmda:
First draft that would practically take hooks and classes under one roof with a shared api, using hooks as backbone. Downsides of course are regressions like "auto". Upsides are 8.x'es refreshed api and new capabilities.


drcmda:
This pull request **introduces 1 alert** when merging c7e924b0b39d5f6fb3274f3e4b529852db86f5ff into 0ebbf23c3a802872d85fa9bdb7354d33ae7065d7 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-8708f99a38c553fd158d63b3ccf12ac5248f60a6)

**new alerts:**

* 1 for Unused variable, import, function or class

---

jacobrask:
I think if we remove auto it would be nice with an official way of doing it, like you recommended d3-ease when removing the easing function. So:

1. A warning when trying to assign `Animated<auto>` to width/height of an `animated` element, with a link to:
2. An easy refactoring path for different cases

drcmda:
@jacobrask Maybe there's still a way, i was thinking about perhaps adding something back like injects or a Globals.use(...) api. When i think of auto today an inject could maybe drop a resize observer (leaving polyfills to the end-user) which can update the owning controller on size changes (or flip to "auto" when needed).

aleclarson:
Merged into #632 🎉

PR #580, opened 2019-03-04
Improve interpolation types

jacobrask:
Also merges the `WithChildren` intermediate classes with `Animated` because nothing extended `Animated` directly without children.


ruijdacd:
In my team, we are currently using the latest version of react-spring (`8.0.27` at the time of writing this) and we the improvements to the interpolation types would be extremely helpful since we are using the `map` and `extrapolate` properties.

Can the improvements made to the types, be pushed to a `8.0.28` version or are these improvements only going to get shipped on `v9`?

By the way, awesome library and keep up the good work :)

PR #584, opened 2019-03-07
Add index arg to `useSprings` `getProps` fn typings

mindblight:
The `useSprings` typings prevent [this](https://codesandbox.io/embed/n9vo1my91p) example. Updated typings to give the index in the callback function.


types/web.d.ts
```diff
@@ -72,6 +72,10 @@ type InferFrom<T extends object> = T extends { to: infer TTo }
 type Merge<A, B> = { [K in keyof A]: K extends keyof B ? B[K] : A[K] } & B
 
 export type SetUpdateFn<DS extends object> = (ds: Partial<UseSpringProps<DS>>) => void
+export interface SetUpdateCallbackFn<DS extends object> {
```
mindblight: Not sure if this sure be somewhere else / more generic.

PR #615, opened 2019-04-01
wip: various improvements

### LINKED ISSUE #335: [useSpring] onStart callback triggered on each frame
Hey! Thanks for this fantastic library.
I'm trying the `useSpring` hook which I understand is experimental, and maybe this is by design or maybe this is me not using `useCallback` or misusing hooks, but I find `onStart` being triggered on each frame.

Here is the isolated codesandbox showing the issue: https://codesandbox.io/s/oj9y6j6z9q
As you can see, dragging the box triggers a `console.log` at each frame.
If you uncomment the commented `onStart` function which increments a state counter, the `onStart` function is run continuously...

### LINKED ISSUE #362: onRest is not called for immediate animations
I would expect either neither onStart or onRest, or both, be called with an immediate animation. Instead only onStart is called. This causes some issues in our application because I want to render the content slightly differently (lower quality images) if we are currently in an active animation. If I `setState({ isAnimating: true })` in `onStart` it will not be reset if the animation is was in `immediate` mode.

    <Spring
      immediate
      from={{ height: 10 }}
      to={{ height: 100 }}
      onStart={() => console.log("start")}
      onRest={() => console.log("rest")}
    >
      {({ height }) => <div style={{ background: "hotpink", height }} />}
    </Spring>

[Edit 8lzp570382](https://codesandbox.io/s/8lzp570382)

### LINKED ISSUE #623: onRest is called again after rerender
Problem:  Once completed, onRest function is called on every rerender.

[minimal codesandbox demo](https://codesandbox.io/s/5x2xxpjoop)

**Expected**:  the count should only increment once.
**Current Behaviour**:  after rerender the onRest function is called again, and the count gets updated.

Versions in this demo:
"react-spring": "**8.0.19**",
"react": "16.8.3",
"react-dom": "16.8.3"

### LINKED ISSUE #624: Sync updates shouldn't cancel pending async updates entirely
Currently, when the `queue` is flushed, any pending async updates are cancelled. 
To ensure that `to` props are never lost, the controller merges them into its `merged` object, which is iterated in the `diff` method.

Merging has the consequence of immediately starting the pending async updates, which I believe is rarely what the user wants. For example, let's say `useTransition` is running `enter` animations on its items along with the `trail` prop (creating a staggered effect). When `useTransition` is then passed an `update`, the `trail` prop is effectively ignored and all `enter` animations begin at the same time.

I think we can do better.

One idea is to attach a timestamp to every flush. Then we never need to cancel anything on flush. Instead, when a pending async update is ready to diff, it checks the timestamp of each animation it wants to update. If the animation has a greater timestamp, skip updating it. By never cancelling on flush, we ensure that end values are never lost. By attaching a timestamp, we ensure that newer updates take precedence over async updates.

Feedback wanted. Thanks!

### LINKED ISSUE #629: Feature Request: different config for multistage transitions
Is it possible to pass different config in next's object parameter? 
```
leave: item => async (next, cancel) => {
    await next({ life: '0%' })
    await next({ opacity: 0 })
    await next({ height: 0 })
  },
```
I really need this functionality to make one step's duration 0ms (for out of screen elements).

If not, I want to make PR request for adding this functionality. Please, let me know.

aleclarson:
I've been doing a ton of work on the `Controller` class today, so I figured I should show you asap, in case you have any concerns or comments. I will explain the changes I've made in a nice list once I'm confident they are working as expected. For now, this is only a draft PR. Cheers!

Closes #335 
Closes #362 
Closes #623 
Closes #624
Closes #629

### DIFF
--- a/src/animated/Controller.ts
+++ b/src/animated/Controller.ts
@@ -4,61 +4,84 @@ import {
   is,
   toArray,
   withDefault,
+  hasKeys,
 } from '../shared/helpers'
 import AnimatedValue from './AnimatedValue'
 import AnimatedValueArray from './AnimatedValueArray'
 import { start, stop } from './FrameLoop'
 import { colorNames, interpolation as interp, now } from './Globals'
+import { SpringProps } from '../../types/renderprops'
 
-type FinishedCallback = (finished?: boolean) => void
+type Animation = any
+
+type ControllerProps<DS extends object> = DS &
+  SpringProps<DS> & {
+    attach?: (ctrl: Controller) => Animation
+  }
 
 type AnimationsFor<P> = { [Key in keyof P]: any }
 
 type ValuesFor<P> = { [Key in keyof P]: any }
 
-type InterpolationsFor<P> = {
-  [Key in keyof P]: P[Key] extends ArrayLike<any>
+type InterpolationsFor<DS> = {
+  [K in keyof DS]: DS[K] extends ArrayLike<any>
     ? AnimatedValueArray
     : AnimatedValue
 }
 
-let G = 0
-class Controller<P extends any = {}> {
-  id: number
+type EndCallback = (finished?: boolean) => void
 
+let nextId = 1
+class Controller<DS extends object = any> {
+  id = nextId++
   idle = true
-  hasChanged = false
   guid = 0
-  local = 0
-  props: P = {} as P
+  props: ControllerProps<DS> = {} as any
   merged: any = {}
-  animations = {} as AnimationsFor<P>
-  interpolations = {} as InterpolationsFor<P>
-  values = {} as ValuesFor<P>
+  animations = {} as AnimationsFor<DS>
+  interpolations = {} as InterpolationsFor<DS>
+  values = {} as ValuesFor<DS>
   configs: any = []
-  listeners: FinishedCallback[] = []
   queue: any[] = []
-  localQueue?: any[]
+  prevQueue: any[] = []
+  onEndQueue: EndCallback[] = []
+  pendingCount = 0
 
-  constructor() {
-    this.id = G++
+  // Add this controller to the frameloop
+  private _start(onEnd?: EndCallback) {
+    if (onEnd) this.onEndQueue.push(onEnd)
+    if (this.idle) {
+      this.idle = false
+      start(this)
+    }
+  }
+
+  private _stop(finished?: boolean) {
+    this.idle = true
+    stop(this)
+    if (finished && this.props.onRest) {
+      this.props.onRest(this.merged)
+    }
+    if (this.onEndQueue.length) {
+      this.onEndQueue.forEach(onEnd => onEnd(finished))
+      this.onEndQueue = []
+    }
   }
 
   /** update(props)
    *  This function filters input props and creates an array of tasks which are executed in .start()
    *  Each task is allowed to carry a delay, which means it can execute asnychroneously */
-  update(args?: P) {
-    //this._id = n + this.id
-
-    if (!args) return this
+  update(args: ControllerProps<DS>) {
     // Extract delay and the to-prop from props
     const { delay = 0, to, ...props } = interpolateTo(args) as any
+
+    // If config is either a function or an array, queue it up as is
     if (is.arr(to) || is.fun(to)) {
-      // If config is either a function or an array queue it up as is
       this.queue.push({ ...props, delay, to })
-    } else if (to) {
-      // Otherwise go through each key since it could be delayed individually
-      let ops: any = {}
+    }
+    // Otherwise go through each key since it could be delayed individually
+    else if (to) {
+      let ops: any[] = []
       Object.entries(to).forEach(([k, v]) => {
         // Fetch delay and create an entry, consisting of the to-props, the delay, and basic props
         const entry = { to: { [k]: v }, delay: callProp(delay, k), ...props }
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
           to: { ...previous, ...entry.to },
         }
       })
-      this.queue = Object.values(ops)
+      ops.forEach(op => this.queue.push(op))
     }
+
     // Sort queue, so that async calls go last
-    this.queue = this.queue.sort((a, b) => a.delay - b.delay)
+    this.queue.sort((a, b) => a.delay - b.delay)
 
     // Diff the reduced props immediately (they'll contain the from-prop and some config)
-    this.diff(props)
+    if (hasKeys(props)) this.diff(props)
+
     return this
   }
 
-  /** start(onEnd)
-   *  This function either executes a queue, if present, or starts the frameloop, which animates */
-  start(onEnd?: FinishedCallback) {
-    // If a queue is present we must excecute it
+  /** start()
+   *  This function either executes a queue, if present, or starts the frameloop, which animates.
+   *  The `useSpring` hooks never have > 1 update per call, because they call this every render. */
+  start(onEnd?: EndCallback) {
+    // If a queue is present we must execute it
     if (this.queue.length) {
-      this.idle = false
+      const { prevQueue } = this
 
       // Updates can interrupt trailing queues, in that case we just merge values
-      if (this.localQueue) {
-        this.localQueue.forEach(({ from = {}, to = {} }) => {
+      if (prevQueue.length) {
+        prevQueue.forEach(({ from, to }) => {
           if (is.obj(from)) this.merged = { ...from, ...this.merged }
           if (is.obj(to)) this.merged = { ...this.merged, ...to }
         })
+        prevQueue.length = 0
       }
 
-      // The guid helps us tracking frames, a new queue over an old one means an override
-      // We discard async calls in that caseÍ
-      const local = (this.local = ++this.guid)
-      const queue = (this.localQueue = this.queue)
-      this.queue = []
+      // The guid helps when tracking frames, a new queue over an old one means an override.
+      // We discard async calls in that case
+      const guid = ++this.guid
+      const queue = (this.prevQueue = this.queue)
+      this.queue = prevQueue
+
+      // Reset delay/async tracking
+      this.pendingCount = 0
 
       // Go through each entry and execute it
-      queue.forEach(({ delay, ...props }, index) => {
-        const cb: FinishedCallback = finished => {
-          if (index === queue.length - 1 && local === this.guid && finished) {
-            this.idle = true
-            if (this.props.onRest) this.props.onRest(this.merged)
-          }
-          if (onEnd) onEnd()
-        }
-
-        // Entries can be delayed, ansyc or immediate
-        let async = is.arr(props.to) || is.fun(props.to)
+      queue.forEach(({ delay, onStart, ...props }) => {
+        // Entries can be delayed, async, or immediate
+        const async = is.arr(props.to) || is.fun(props.to)
         if (delay) {
+          this.pendingCount++
           setTimeout(() => {
-            if (local === this.guid) {
-              if (async) this.runAsync(props, cb)
-              else this.diff(props).start(cb)
+            if (guid === this.guid) {
+              this.pendingCount--
+              if (async) this.runAsync(guid, props, onEnd)
+              else this.diff(props)._start(onEnd)
             }
           }, delay)
-        } else if (async) this.runAsync(props, cb)
-        else this.diff(props).start(cb)
+        } else {
+          if (async) this.runAsync(guid, props, onEnd)
+          else this.diff(props)._start(onEnd)
+        }
       })
     }
-    // Otherwise we kick of the frameloop
-    else {
-      if (is.fun(onEnd)) this.listeners.push(onEnd)
-      if (this.props.onStart) this.props.onStart()
-      start(this)
-    }
+    // Otherwise ensure the frameloop is active
+    else this._start(onEnd)
     return this
   }
 
   stop(finished?: boolean) {
-    this.listeners.forEach(onEnd => onEnd(finished))
-    this.listeners = []
+    if (!this.idle || this.pendingCount) {
+      this.guid++
+      this._stop(finished)
+      this.pendingCount = 0
+    }
     return this
   }
 
-  /** Pause sets onEnd listeners free, but also removes the controller from the frameloop */
-  pause(finished?: boolean) {
-    this.stop(true)
-    if (finished) stop(this)
-    return this
+  // Called by the frameloop
+  onFrame(isActive: boolean) {
+    if (this.props.onFrame) {
+      this.props.onFrame(this.values)
+    }
+    if (!isActive) {
+      this._stop(true)
+    }
   }
 
-  runAsync({ delay, ...props }: P, onEnd: FinishedCallback) {
-    const local = this.local
+  runAsync(
+    guid: number,
+    props: SpringProps<DS>,
+    onEnd?: (finished: boolean) => void
+  ) {
     // If "to" is either a function or an array it will be processed async, therefor "to" should be empty right now
     // If the view relies on certain values "from" has to be present
-    let queue = Promise.resolve(undefined)
-    if (is.arr(props.to)) {
-      for (let i = 0; i < props.to.length; i++) {
-        const index = i
-        const fresh = { ...props, ...interpolateTo(props.to[index]) }
-        if (is.arr(fresh.config)) fresh.config = fresh.config[index]
-        queue = queue.then(
-          (): Promise<any> | void => {
-            //this.stop()
-            if (local === this.guid)
-              return new Promise(r => this.diff(fresh).start(r))
-          }
-        )
-      }
-    } else if (is.fun(props.to)) {
-      let index = 0
+    let queue = Promise.resolve()
+
+    const { to } = props
+    if (is.arr(to)) {
+      to.forEach((p, i) => {
+        queue = queue.then(() => {
+          if (guid !== this.guid) return
+          const fresh = { ...props, ...interpolateTo(p) }
+          if (is.arr(fresh.config)) fresh.config = fresh.config[i]
+          return new Promise<any>(r => this.diff(fresh).start(r))
+        })
+      })
+    } else if (is.fun(to)) {
+      let i = 0
       let last: Promise<any>
       queue = queue.then(() =>
-        props
-          .to(
-            // next(props)
-            (p: P) => {
-              const fresh = { ...props, ...interpolateTo(p) }
-              if (is.arr(fresh.config)) fresh.config = fresh.config[index]
-              index++
-              //this.stop()
-              if (local === this.guid)
-                return (last = new Promise(r => this.diff(fresh).start(r)))
-              return
-            },
-            // cancel()
-            (finished = true) => this.stop(finished)
-          )
-          .then(() => last)
+        to(
+          // next(props)
+          p => {
+            if (guid !== this.guid) return
+            const fresh = { ...props, ...interpolateTo(p) }
+            if (is.arr(fresh.config)) fresh.config = fresh.config[i++]
+            return (last = new Promise(r => this.diff(fresh).start(r)))
+          },
+          // cancel()
+          finished => this.stop(finished)
+        ).then(() => last)
       )
     }
-    queue.then(onEnd)
+
+    this.pendingCount++
+    queue.then(() => {
+      if (guid === this.guid) {
+        this.pendingCount--
+        if (onEnd) onEnd(true)
+      }
+    })
   }
 
   diff(props: any) {
@@ -208,9 +239,12 @@ class Controller<P extends any = {}> {
     // This will collect all props that were ever set, reset merged props when necessary
     this.merged = { ...from, ...this.merged, ...to }
 
-    this.hasChanged = false
     // Attachment handling, trailed springs can "attach" themselves to a previous spring
     let target = attach && attach(this)
+
+    // Detect when no animations are changed
+    let changed = false
+
     // Reduces input { name: value } pairs into animated values
     this.animations = Object.entries<any>(this.merged).reduce(
       (acc, [name, value]) => {
@@ -287,7 +321,6 @@ class Controller<P extends any = {}> {
           animatedValues = toArray(parent.getPayload())
           if (reset && !isInterpolation) parent.setValue(fromValue, false)
 
-          this.hasChanged = true
           // Reset animated values
           animatedValues.forEach(value => {
             value.startPosition = value.value
@@ -304,56 +337,52 @@ class Controller<P extends any = {}> {
             parent.setValue(isInterpolation ? toValue : value, false)
           }
 
-          return {
-            ...acc,
-            [name]: {
-              ...entry,
-              name,
-              parent,
-              interpolation,
-              animatedValues,
-              toValues,
-              previous: newValue,
-              config: toConfig,
-              fromValues: toArray(parent.getValue()),
-              immediate: callProp(immediate, name),
-              initialVelocity: withDefault(toConfig.velocity, 0),
-              clamp: withDefault(toConfig.clamp, false),
-              precision: withDefault(toConfig.precision, 0.01),
-              tension: withDefault(toConfig.tension, 170),
-              friction: withDefault(toConfig.friction, 26),
-              mass: withDefault(toConfig.mass, 1),
-              duration: toConfig.duration,
-              easing: withDefault(toConfig.easing, (t: number) => t),
-              decay: toConfig.decay,
-            },
+          changed = true
+          acc[name] = {
+            ...entry,
+            name,
+            parent,
+            interpolation,
+            animatedValues,
+            toValues,
+            previous: newValue,
+            config: toConfig,
+            fromValues: toArray(parent.getValue()),
+            immediate: callProp(immediate, name),
+            initialVelocity: withDefault(toConfig.velocity, 0),
+            clamp: withDefault(toConfig.clamp, false),
+            precision: withDefault(toConfig.precision, 0.01),
+            tension: withDefault(toConfig.tension, 170),
+            friction: withDefault(toConfig.friction, 26),
+            mass: withDefault(toConfig.mass, 1),
+            duration: toConfig.duration,
+            easing: withDefault(toConfig.easing, (t: number) => t),
+            decay: toConfig.decay,
           }
-        } else {
-          if (!currentValueDiffersFromGoal) {
-            // So ... the current target value (newValue) appears to be different from the previous value,
-            // which normally constitutes an update, but the actual value (currentValue) matches the target!
-            // In order to resolve this without causing an animation update we silently flag the animation as done,
-            // which it technically is. Interpolations also needs a config update with their target set to 1.
-            if (isInterpolation) {
-              parent.setValue(1, false)
-              interpolation.updateConfig({ output: [newValue, newValue] })
-            }
+        } else if (!currentValueDiffersFromGoal) {
+          // So ... the current target value (newValue) appears to be different from the previous value,
+          // which normally constitutes an update, but the actual value (currentValue) matches the target!
+          // In order to resolve this without causing an animation update we silently flag the animation as done,
+          // which it technically is. Interpolations also needs a config update with their target set to 1.
+          if (isInterpolation) {
+            parent.setValue(1, false)
+            interpolation.updateConfig({ output: [newValue, newValue] })
+          }
 
-            parent.done = true
-            this.hasChanged = true
-            return { ...acc, [name]: { ...acc[name], previous: newValue } }
-          }
-          return acc
+          changed = true
+          parent.done = true
+          acc[name] = { ...entry, previous: newValue }
         }
+        return acc
       },
       this.animations
     )
 
-    if (this.hasChanged) {
+    if (changed) {
       // Make animations available to frameloop
       this.configs = Object.values(this.animations)
-      this.values = {} as ValuesFor<P>
-      this.interpolations = {} as InterpolationsFor<P>
+      this.values = {} as ValuesFor<DS>
+      this.interpolations = {} as InterpolationsFor<DS>
       for (let key in this.animations) {
         this.interpolations[key] = this.animations[key].interpolation
         this.values[key] = this.animations[key].interpolation.getValue()
@@ -364,13 +393,12 @@ class Controller<P extends any = {}> {
 
   destroy() {
     this.stop()
-    this.props = {} as P
+    this.props = {} as DS
     this.merged = {}
-    this.animations = {} as AnimationsFor<P>
-    this.interpolations = {} as InterpolationsFor<P>
-    this.values = {} as ValuesFor<P>
+    this.animations = {} as AnimationsFor<DS>
+    this.interpolations = {} as InterpolationsFor<DS>
+    this.values = {} as ValuesFor<DS>
     this.configs = []
-    this.local = 0
   }
 
   getValues = () => this.interpolations
--- a/src/animated/Controller.ts
+++ b/src/animated/Controller.ts
@@ -1,10 +1,11 @@
 import {
   callProp,
+  fillArray,
+  hasKeys,
   interpolateTo,
   is,
   toArray,
   withDefault,
-  hasKeys,
 } from '../shared/helpers'
 import AnimatedValue from './AnimatedValue'
 import AnimatedValueArray from './AnimatedValueArray'
@@ -89,13 +90,13 @@ class Controller<DS extends object = any> {
         // Eagerly create an Animated node to be used during render.
         // When `from[name]` exists, the node is created by `_diff` instead.
         if (!hasFrom && !this.values[name]) {
+          // Use undefined as the placeholder value
+          const nil = undefined as any
           const node = is.arr(value)
-            ? new AnimatedValueArray(
-                new Array(value.length).fill(() => new AnimatedValue(NaN))
-              )
+            ? new AnimatedValueArray(fillArray(value.length, () => nil))
             : isAnimatableString(value)
-            ? new AnimatedValue(NaN).interpolate({ output: [] })
-            : new AnimatedValue(NaN)
+            ? new AnimatedValue(nil).interpolate({ output: [] })
+            : new AnimatedValue(nil)
 
           // For `node` to be reused, an animation object must exist
           const animation: any = {
@@ -241,16 +242,27 @@ class Controller<DS extends object = any> {
   }
 
   private _run(guid: number, props: UpdateProps<DS>, onEnd?: OnEnd) {
-    // Never call `onStart` for immediate animations
-    if (!props.immediate) {
-      const { onStart } = props
-      // Allow `useCallback` to prevent multiple calls
-      if (onStart && onStart !== this.props.onStart) onStart()
-    }
     if (is.arr(props.to) || is.fun(props.to)) {
       this._runAsync(guid, props, onEnd)
     } else {
       this._diff(props)._start(onEnd)
+    }
+  }
+
+  private _willStart(props: SpringProps<any>, oldProps: SpringProps<any>) {
+    const { immediate, to, onStart } = props
+
+    // Equals true when all animations are immediate
+    const isImmediate =
+      immediate === true ||
+      (is.fun(immediate) && Object.keys(to).every(name => immediate(name)))
+
+    // Never call `onStart` for immediate animations
+    if (!isImmediate) {
+      // Allow `useCallback` to prevent multiple calls
+      if (onStart && onStart !== oldProps.onStart) {
+        onStart(props)
+      }
     }
   }
 
@@ -297,7 +309,8 @@ class Controller<DS extends object = any> {
   }
 
   private _diff(props: UpdateProps<DS>) {
-    this.props = { ...this.props, ...props }
+    const oldProps = this.props
+    this.props = { ...oldProps, ...props }
     let {
       from = {} as any,
       to = {} as any,
@@ -437,6 +450,7 @@ class Controller<DS extends object = any> {
     )
 
     if (changed) {
+      this._willStart(props, oldProps)
       const frames = (this.frames = {} as any)
       const values = (this.values = {} as any)
       for (const key in this.animations) {
--- a/src/animated/Controller.ts
+++ b/src/animated/Controller.ts
@@ -56,7 +56,7 @@ class Controller<DS extends object = any> {
   queue: any[] = []
   prevQueue: any[] = []
   onEndQueue: OnEnd[] = []
-  pendingCount = 0
+  runCount = 0
 
   getValues = () => this.values as { [K in keyof DS]: Animation<DS[K]>['node'] }
 
@@ -65,6 +65,8 @@ class Controller<DS extends object = any> {
    * Individual tasks may be async and/or delayed.
    */
   update(updateProps: UpdateProps<DS>) {
+    if (!updateProps) return this
+
     // Extract delay and the to-prop from props
     const { delay = 0, to, ...props } = interpolateTo(updateProps) as any
 
@@ -146,7 +148,6 @@ class Controller<DS extends object = any> {
 
         // Reset any queue-related state
         prevQueue.length = 0
-        this.pendingCount = 0
         this.onEndQueue.length = 0
       }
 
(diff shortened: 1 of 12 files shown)


src/animated/Controller.ts
```diff
@@ -4,61 +4,84 @@ import {
...
+    attach?: (ctrl: Controller) => Animation
+  }
 
 type AnimationsFor<P> = { [Key in keyof P]: any }
 
 type ValuesFor<P> = { [Key in keyof P]: any }
 
-type InterpolationsFor<P> = {
-  [Key in keyof P]: P[Key] extends ArrayLike<any>
+type InterpolationsFor<DS> = {
```
aleclarson: I'm not actually sure what `DS` stands for, but it's being used in the `SpringProps` type, so I wanted to match with that.

src/animated/Controller.ts
```diff
@@ -4,61 +4,84 @@ import {
...
-  id: number
+type EndCallback = (finished?: boolean) => void
 
+let nextId = 1
+class Controller<DS extends object = any> {
+  id = nextId++
   idle = true
-  hasChanged = false
   guid = 0
-  local = 0
```
aleclarson: I dropped the `local` property in favor of passing the `guid` through the `runAsync` method, since that's the only place the `local` property was being used.

src/animated/Controller.ts
```diff
@@ -4,61 +4,84 @@ import {
...
-let G = 0
-class Controller<P extends any = {}> {
-  id: number
+type EndCallback = (finished?: boolean) => void
 
+let nextId = 1
+class Controller<DS extends object = any> {
+  id = nextId++
   idle = true
-  hasChanged = false
```
aleclarson: Since this property is only referenced in the `diff` method, it makes more sense as a local variable.

src/animated/FrameLoop.ts
```diff
@@ -137,7 +132,7 @@ const update = () => {
 }
 
 const start = (controller: Controller) => {
-  if (!controllers.has(controller)) controllers.add(controller)
+  controllers.add(controller)
```
aleclarson: The `has` call is unnecessary since `controllers` is a `Set` object.

src/animated/FrameLoop.ts
```diff
@@ -146,7 +141,7 @@ const start = (controller: Controller) => {
 }
 
 const stop = (controller: Controller) => {
-  if (controllers.has(controller)) controllers.delete(controller)
+  controllers.delete(controller)
```
aleclarson: Same for the `delete` method.

src/useSprings.js
```diff
@@ -25,9 +25,7 @@ export const useSprings = (length, props) => {
         const ctrl = new Ctrl()
         const newProps = isFn ? callProp(props, i, ctrl) : props[i]
         if (i === 0) ref = newProps.ref
-        ctrl.update(newProps)
-        if (!ref) ctrl.start()
```
aleclarson: I think the `start` method should only be called within `useEffect`, to avoid starting the animation(s) before we know if the render was interrupted (which will be possible when Suspense is released). Am I misguided here?

src/animated/Controller.ts
```diff
@@ -287,7 +321,6 @@ class Controller<P extends any = {}> {
           animatedValues = toArray(parent.getPayload())
           if (reset && !isInterpolation) parent.setValue(fromValue, false)
 
-          this.hasChanged = true
```
aleclarson: Strangely, this was being "initialized" to `true` instead of `false`. 😅

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
           to: { ...previous, ...entry.to },
         }
       })
-      this.queue = Object.values(ops)
```
aleclarson: 😳 The `update` method should never replace the `queue`. So I converted the `ops` object to a sparse array!

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
+        to(
+          // next(props)
+          p => {
+            if (guid !== this.guid) return
+            const fresh = { ...props, ...interpolateTo(p) }
+            if (is.arr(fresh.config)) fresh.config = fresh.config[i++]
+            return (last = new Promise(r => this.diff(fresh).start(r)))
+          },
+          // cancel()
+          finished => this.stop(finished)
```
aleclarson: The `finished` argument used to default to `true`, but that didn't make sense to me. Usually, if you're calling this function, it's to signal that the animation is cancelled. To signal that the animation finished cleanly, just return without calling either of these functions.

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
-              return
-            },
-            // cancel()
-            (finished = true) => this.stop(finished)
-          )
-          .then(() => last)
+        to(
+          // next(props)
+          p => {
+            if (guid !== this.guid) return
```
aleclarson: This now compares `guid` values *before* creating "fresh" props.

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
+            const fresh = { ...props, ...interpolateTo(p) }
+            if (is.arr(fresh.config)) fresh.config = fresh.config[i++]
+            return (last = new Promise(r => this.diff(fresh).start(r)))
+          },
+          // cancel()
+          finished => this.stop(finished)
+        ).then(() => last)
       )
     }
-    queue.then(onEnd)
```
aleclarson: We really shouldn't be calling `onEnd` without comparing `guid` values first.

src/animated/Controller.ts
```diff
@@ -304,56 +337,52 @@ class Controller<P extends any = {}> {
...
-              precision: withDefault(toConfig.precision, 0.01),
-              tension: withDefault(toConfig.tension, 170),
-              friction: withDefault(toConfig.friction, 26),
-              mass: withDefault(toConfig.mass, 1),
-              duration: toConfig.duration,
-              easing: withDefault(toConfig.easing, (t: number) => t),
-              decay: toConfig.decay,
-            },
+          changed = true
+          acc[name] = {
```
aleclarson: I think it's better to avoid creating a new copy of `acc` for every changed animation, so I switched this to a simple assignment. WDYT?

src/animated/Controller.ts
```diff
@@ -4,61 +4,84 @@ import {
...
   configs: any = []
-  listeners: FinishedCallback[] = []
   queue: any[] = []
-  localQueue?: any[]
+  prevQueue: any[] = []
+  onEndQueue: EndCallback[] = []
+  pendingCount = 0
+
+  // Add this controller to the frameloop
+  private _start(onEnd?: EndCallback) {
```
aleclarson: Do you like the private methods above or below the public methods?

src/useSprings.js
drcmda: you may be right, putting it into useEffect should be fine

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
+      ops.forEach(op => this.queue.push(op))
     }
+
     // Sort queue, so that async calls go last
-    this.queue = this.queue.sort((a, b) => a.delay - b.delay)
+    this.queue.sort((a, b) => a.delay - b.delay)
 
     // Diff the reduced props immediately (they'll contain the from-prop and some config)
-    this.diff(props)
+    if (hasKeys(props)) this.diff(props)
```
aleclarson: Added a fast path for when only `to` and `delay` are defined in `props`.

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
           to: { ...previous, ...entry.to },
         }
       })
-      this.queue = Object.values(ops)
+      ops.forEach(op => this.queue.push(op))
     }
+
     // Sort queue, so that async calls go last
-    this.queue = this.queue.sort((a, b) => a.delay - b.delay)
```
aleclarson: `Array.prototype.sort` sorts "in place", so no assignment necessary! :)

src/animated/Controller.ts
```diff
@@ -4,61 +4,84 @@ import {
...
+    }
   }
 
   /** update(props)
    *  This function filters input props and creates an array of tasks which are executed in .start()
    *  Each task is allowed to carry a delay, which means it can execute asnychroneously */
-  update(args?: P) {
-    //this._id = n + this.id
-
-    if (!args) return this
```
aleclarson: I didn't see anywhere in the codebase that called `update` with an argument that's potentially falsy, so I removed this line. LMK if I should revert.

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
+   *  The `useSpring` hooks never have > 1 update per call, because they call this every render. */
+  start(onEnd?: EndCallback) {
+    // If a queue is present we must execute it
     if (this.queue.length) {
-      this.idle = false
+      const { prevQueue } = this
 
       // Updates can interrupt trailing queues, in that case we just merge values
-      if (this.localQueue) {
-        this.localQueue.forEach(({ from = {}, to = {} }) => {
```
aleclarson: I removed the `= {}` parts, because duck-typing is used via the `is.obj` function.

src/animated/Controller.ts
```diff
@@ -69,123 +92,131 @@ class Controller<P extends any = {}> {
...
+      this.queue = prevQueue
 
-      // Go through each entry and execute it
-      queue.forEach(({ delay, ...props }, index) => {
-        const cb: FinishedCallback = finished => {
-          if (index === queue.length - 1 && local === this.guid && finished) {
-            this.idle = true
-            if (this.props.onRest) this.props.onRest(this.merged)
-          }
-          if (onEnd) onEnd()
```
aleclarson: The `onEnd` function wasn't being passed the `finished` value, so I fixed that.

src/animated/Controller.ts
```diff
@@ -127,25 +100,33 @@ class Controller<DS extends object = any> {
...
       this.queue = prevQueue
 
-      // Reset delay/async tracking
+      // Reset any queue-related state
       this.pendingCount = 0
+      this.onEndQueue.length = 0
+
+      // Never assume that the last update always finishes last, since that's
+      // not true when 2+ async animations have indeterminate durations.
+      let remaining = queue.length
```
aleclarson: The controller no longer assumes that the last "update" will always be the last to finish. 🎉

src/animated/Controller.ts
```diff
@@ -163,21 +154,52 @@ class Controller<DS extends object = any> {
...
+    }
+    if (this.onEndQueue.length) {
+      this.onEndQueue.forEach(onEnd => onEnd(finished))
+      this.onEndQueue = []
+    }
+  }
+
+  private _run(guid: number, props: UpdateProps<DS>, onEnd?: OnEnd) {
+    // Never call `onStart` for immediate animations
+    if (!props.immediate) {
```
aleclarson: This fixes #362 by not calling `onStart` for immediate animations.

src/animated/Controller.ts
```diff
@@ -163,21 +154,52 @@ class Controller<DS extends object = any> {
...
+      const { onStart } = props
+      // Allow `useCallback` to prevent multiple calls
+      if (onStart && onStart !== this.props.onStart) onStart()
```
aleclarson: This fixes #335 by not calling `onStart` when it's wrapped with `useCallback`.

src/animated/Controller.ts
```diff
@@ -159,17 +165,21 @@ class Controller<DS extends object = any> {
...
-      this.idle = false
-      start(this)
+    if (this.configs.length) {
+      if (onEnd) this.onEndQueue.push(onEnd)
+      if (this.idle) {
+        this.idle = false
+        start(this)
+      }
+    } else if (onEnd) {
+      onEnd(true)
```
aleclarson: The `start` method now calls its `onEnd` argument immediately if no animations exist.

Idk if this edge case matters, but I decided to put it in for posterity.

src/useSprings.js
```diff
@@ -13,52 +13,48 @@ export const useSprings = (length, props) => {
...
+          return ctrl.current
+        },
+      },
     ]
   }, [length])
 
-  ctrl.current = controllers
-
-  // The hooks reference api gets defined here ...
-  const api = useImperativeHandle(ref, () => ({
```
aleclarson: The `useImperativeHandle` hook has a `deps` argument. Idk if omitting the `deps` argument causes the `api` to be recreated on every render, but I moved the `api` into the `useMemo` call above (just to be safe).

drcmda:
This pull request **fixes 1 alert** when merging 7113be6ec538657a055c4823867dfb822231b1b0 into be6dd743a0900e4f3ede325e0fb30cd65282de92 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-b9a5bb9533f06c2d41c01b4ec1bfbce60ee39305)

**fixed alerts:**

* 1 for Unused variable, import, function or class

---

src/useSprings.js
```diff
@@ -13,52 +13,48 @@ export const useSprings = (length, props) => {
...
+          Promise.all(ctrl.current.map(c => new Promise(r => c.start(r)))),
+        stop: finished => ctrl.current.forEach(c => c.stop(finished)),
+        get controllers() {
+          return ctrl.current
+        },
+      },
     ]
   }, [length])
 
-  ctrl.current = controllers
```
aleclarson: I moved this into the `useEffect` call that runs on every render, since you're not supposed to mutate stuff during render ([see here](https://twitter.com/dan_abramov/status/1102182683885617159)).

src/useSprings.js
```diff
@@ -13,52 +13,48 @@ export const useSprings = (length, props) => {
...
-    start: () =>
-      Promise.all(ctrl.current.map(c => new Promise(r => c.start(r)))),
-    stop: finished => ctrl.current.forEach(c => c.stop(finished)),
-    get controllers() {
-      return ctrl.current
-    },
-  }))
-
-  // This function updates the controllers
-  const updateCtrl = useMemo(
```
aleclarson: I moved this into the first `useMemo` call, since they have the same deps.

src/useSprings.js
```diff
@@ -13,52 +13,48 @@ export const useSprings = (length, props) => {
...
-    () => updateProps =>
-      ctrl.current.map((c, i) => {
-        c.update(isFn ? callProp(updateProps, i, c) : updateProps[i])
```
aleclarson: This was incorrectly using `isFn` instead of `is.fun(updateProps)`, so I fixed it.

src/useSprings.js
```diff
@@ -13,52 +13,48 @@ export const useSprings = (length, props) => {
...
-      ctrl.current = undefined
-    }
+  const [controllers, setProps, ref, api] = useMemo(() => {
     let ref
     return [
+      // Recreate the controllers whenever `length` changes
       new Array(length).fill().map((_, i) => {
-        const ctrl = new Ctrl()
-        const newProps = isFn ? callProp(props, i, ctrl) : props[i]
+        const c = new Ctrl()
```
aleclarson: This variable was shadowing another, so I renamed it.

src/animated/Controller.ts
drcmda: @jacobrask @Jessidhia

src/animated/Controller.ts
drcmda: local took the current guid so that async springs can bail out if their frame is out of date, i guess passing it through is the same thing

src/animated/Controller.ts
drcmda: the question is what constitutes an end, it could be merely that listeners are fired, or that the controller is ripped out of the frameloop. both have usecases with side effects in some specific situations.

src/animated/Controller.ts
drcmda: looks good!

src/animated/Controller.ts
drcmda: that was just for debug purposes

drcmda:
This looks fantastic!

if you want to stresstest queues, the most terrible of all usecases has been this one: https://github.com/react-spring/react-spring-examples/tree/master/demos/tests/freshtilledsoil

```jsx
import examples from './components/examples-tests'

//const DEBUG = false
const DEBUG = 'fresh'
```

src/animated/Controller.ts
aleclarson: I think of an "animation end" as (1) a spring/decay coming to rest, (2) a duration hitting zero, (3) the last keyframe ending, or (4) an async animator returning undefined. The frameloop should be an implementation detail whenever possible. Does that clear things up?

src/useSprings.js
aleclarson: Looks like we'll need to create the `Animated` objects eagerly since we're not calling `start` immediately. I guess we could call the `diff` method for all synchronous "updates". I'll test that out and report back with my findings.

src/animated/Controller.ts
aleclarson: I just realized that `immediate` can be a function, so the presence of `props.immediate` is not enough to know whether `onStart` should be called.

drcmda: oh, right, theres a helper for it, callProp, it's used throughout the lib

drcmda:
This pull request **fixes 2 alerts** when merging 6fdacbcc77a7040191f33f4a1d631c4b357eb10a into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-e5c8c8ef256b7ebc9267414a051b08966ad444ea)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 570ff2d985a2d26e6cd7a526b34c04c45394b41d into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-8053422e987b84803bb0cdbc1e3866b4a93eb568)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging e40a463c15f2fba3038308e13cfd1ff6815d3d1f into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-9112741f0f5524b0f643cd261782cdd2540c3d1e)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -281,129 +296,114 @@ class Controller<DS extends object = any> {
...
-            toValues,
-            previous: newValue,
-            config: toConfig,
-            fromValues: toArray(parent.getValue()),
-            immediate: callProp(immediate, name),
+            node,
+            immediate: isImmediate,
+            goalValue,
+            toValues: toArray(target ? toValue.getPayload() : toValue),
+            fromValues: toArray(node.getValue()),
```
aleclarson: The `toValues` and `fromValues` are now calculated here!

drcmda:
This pull request **fixes 2 alerts** when merging dc90e508e86a5b4ef16cc71595cd3afb6c2131eb into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-17387056a798677634fd5762280167e47807bced)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -281,129 +296,114 @@ class Controller<DS extends object = any> {
...
+            animatedValues:
```
aleclarson: The `toValues` and `animatedValues` are now calculated directly on the animation object.

src/animated/Controller.ts
```diff
@@ -281,129 +296,111 @@ class Controller<DS extends object = any> {
...
-            previous: newValue,
-            config: toConfig,
-            fromValues: toArray(parent.getValue()),
-            immediate: callProp(immediate, name),
+            node,
+            immediate: isImmediate,
+            goalValue,
+            toValues: toArray(target ? toValue.getPayload() : toValue),
+            fromValues: toArray(node.getValue()),
+            animatedValues: toArray(node.getPayload() as any),
```
aleclarson: The `toValues` and `animatedValues` are now calculated directly on the animation object.

drcmda:
This pull request **fixes 2 alerts** when merging 584b4b0efc730f26fe60c3c0cc7d017084dc259b into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-c0441116414cd99b09e1b842bcda3bca6f179285)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

aleclarson:
This part doesn't seem right:

https://github.com/react-spring/react-spring/blob/bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2/src/animated/Controller.ts#L89-L95

That doesn't seem to account for "updates" with unfinished delays.

aleclarson:
### Random notes

Note: `reset` must be used whenever the `attach` target changes. otherwise, only changed animations are attached. this bug may affect useTrail (like when the `reverse` prop changes).

Note: using `attach` with `immediate` on an interpolated string ends up passing an Animated node to the `setValue` method of AnimatedValue. not sure if that's an edge case we care about, but there it is. :P

aleclarson:
Q: Why do `this.merged` values take precedence over `from` values in the `diff` method?

https://github.com/react-spring/react-spring/blob/bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2/src/animated/Controller.ts#L209

Further down in this function, the `from` object is still being accessed for the `fromValue`, so this confused me a little.

drcmda:
This pull request **fixes 2 alerts** when merging 1c2278a11d48213867a140bcad42fad7fb9a40a7 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-20108d3c28baa6fc3c8add339237c2aa0465eb17)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 7af349dd9d02dc077c827123bea9fda5ec2e5838 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-e752ec732b38dbb6363ba96c109dd2d317f117d5)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 9dcb34e922901eb8d083e50fd99c9825ba64b07b into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-1fe2b6c7d99b34c204e0d480c6f471fdedb8e77f)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -437,6 +450,7 @@ class Controller<DS extends object = any> {
     )
 
     if (changed) {
+      this._willStart(props, oldProps)
```
aleclarson: The `onStart` prop is now called for every "update" that adds or changes an animation. Except when all animations are immediate.

src/animated/Controller.ts
```diff
@@ -241,19 +242,30 @@ class Controller<DS extends object = any> {
...
-      if (onStart && onStart !== this.props.onStart) onStart()
-    }
     if (is.arr(props.to) || is.fun(props.to)) {
       this._runAsync(guid, props, onEnd)
     } else {
       this._diff(props)._start(onEnd)
     }
   }
 
+  private _willStart(props: SpringProps<any>, oldProps: SpringProps<any>) {
```
aleclarson: Note: The `props` argument is only the "update" props. 
Not the new value of `this.props`.

src/animated/Controller.ts
```diff
@@ -241,19 +242,30 @@ class Controller<DS extends object = any> {
...
+    // Equals true when all animations are immediate
+    const isImmediate =
+      immediate === true ||
+      (is.fun(immediate) && Object.keys(to).every(name => immediate(name)))
+
+    // Never call `onStart` for immediate animations
+    if (!isImmediate) {
+      // Allow `useCallback` to prevent multiple calls
+      if (onStart && onStart !== oldProps.onStart) {
+        onStart(props)
```
aleclarson: Pass the "update" props to the `onStart` prop.

drcmda:
This pull request **fixes 2 alerts** when merging 74ba20befc84d6b534d337d15ff9d773e7d73b68 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-d5c6dac93b01a3a81817745f31768c0e0a3fb8cf)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -90,13 +89,11 @@ class Controller<DS extends object = any> {
...
-          // Use undefined as the placeholder value
-          const nil = undefined as any
           const node = is.arr(value)
-            ? new AnimatedValueArray(fillArray(value.length, () => nil))
+            ? new AnimatedValueArray(value)
             : isAnimatableString(value)
-            ? new AnimatedValue(nil).interpolate({ output: [] })
-            : new AnimatedValue(nil)
+            ? new AnimatedValue(0).interpolate({ output: [value, value] })
+            : new AnimatedValue(value as any)
```
aleclarson: No longer using a placeholder value like `undefined` or `NaN`.

Instead, the `to[name]` value is used as the initial value when `from[name]` is undefined.

drcmda:
This pull request **fixes 2 alerts** when merging c0dbf0d748447f6c05b6cac74cf742ccce535743 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-62d08fdc6dbfb225e2a966ba1c66f3fb3c11f3a8)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -386,22 +383,22 @@ class Controller<DS extends object = any> {
...
+              input = animatedValues[0]
 
-              if (!reset) output[0] = node.calc(parent.value)
+              if (!reset) output[0] = node.calc(input.value)
               node.updateConfig({ output })
 
-              parent.reset(isActive)
-              parent.setValue(0, false)
+              input.setValue(0, false)
+              input.reset(isActive)
```
aleclarson: Note: It's very important to call `setValue` before `reset` here.

src/animated/Controller.ts
aleclarson: No longer using a placeholder value like `undefined` or `NaN`.

Instead, the `to[name]` value is used as the initial value when `from[name]` is undefined.

src/animated/Controller.ts
aleclarson: Note to self: It's very important to call `setValue` before `reset` here.

src/useSprings.js
```diff
@@ -24,11 +24,14 @@ export const useSprings = (length, props) => {
         return c.update(newProps)
       }),
       // This updates the controllers with new props
-      props =>
+      props => {
+        const isFn = is.fun(props)
+        if (!isFn) props = toArray(props)
```
aleclarson: Previously, this function would only take an array or a function. Now you can pass an object without `Controller#update` receiving an undefined value on accident.

This affected `useSpring` in particular.

src/animated/Controller.ts
```diff
@@ -306,8 +291,7 @@ class Controller<DS extends object = any> {
   }
 
   private _diff(props: UpdateProps<DS>) {
-    const oldProps = this.props
-    this.props = { ...oldProps, ...props }
+    Object.assign(this.props, props)
```
aleclarson: Avoid cloning `this.props` on every diff.

drcmda:
This pull request **fixes 2 alerts** when merging 9e2dfe1b7916e95dcb46c5f5e869a8d769e4bb16 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-f671d02c94258c91bf9b7bc9fa5859ebcddd19e8)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -450,14 +439,18 @@ class Controller<DS extends object = any> {
     )
 
     if (changed) {
-      this._willStart(props, oldProps)
+      if (started.length && onStart) {
+        started.forEach(name => onStart!(this.animations[name]))
```
aleclarson: Call `onStart` once per animated key, whenever a key has a new goal value and is not immediate. @drcmda Does this sound good?

This definitely messes with `useTransition`, so I've gotta figure out a workaround there.

drcmda:
This pull request **fixes 2 alerts** when merging 9b438ce7776433936215e277649af11e3d1e23c1 into bdf8afb8b2d837fd756f8ffaae6931fb15c0e3d2 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-1a0a00251b03c0be3f89e2ae08059158109b1702)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

types/renderprops-universal.d.ts
```diff
@@ -172,15 +174,22 @@ export interface TransitionProps<
...
    */
   update?: TUpdate | ((item: TItem) => TUpdate)
+  /**
+   * Called when an item's transition will begin
+   */
+  onStart?: (item: TItem, phase: TransitionPhase) => void
+  /**
+   * Called when an item's transition has come to a stand-still
+   */
+  onRest?: (item: TItem, phase: TransitionPhase, restValues: DS) => void
```
aleclarson: Fixed the argument types of the `onStart`, `onRest`, and `onFrame` props in `TransitionProps`.

src/animated/Controller.ts
aleclarson: Perhaps this would be better as a new `onAnimation` prop.

Then I'll keep `onStart` the same, where it's only called when the first animation of an "update" is about to start.

aleclarson: Actually, I think the utility of `onStart`'s current behavior is dubious. You can always throttle `onStart` if you only want it called once per update, which is what `useTransition` will do.

drcmda:
This pull request **fixes 2 alerts** when merging 7d03f2cb93a566f522897957ec1a4197cbe4431a into 46a357e89e8b8360cd1b581078a8c1dae043f33e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-aa07d0cc8ae3524522d9e981897cfbd62b8fdd3a)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 19bdbfefb9286c2a5d791a9ae8f4ec809ec18ca4 into 46a357e89e8b8360cd1b581078a8c1dae043f33e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-122933766ce4bb5795e3f5457adc0d9541221d12)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 2bbe9ad26f88e11e23c469310fae8dc89dfc1fe3 into 46a357e89e8b8360cd1b581078a8c1dae043f33e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-d55a304fea8027a6fe134db2ce820200cc7fbb31)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -238,13 +183,73 @@ class Controller<DS extends object = any> {
...
+    const { onEndQueue } = this
+    if (onEndQueue.length) {
       this.onEndQueue = []
-      queue.forEach(onEnd => onEnd(finished))
+      onEndQueue.forEach(onEnd => onEnd(finished))
+    }
+  }
+
+  // Execute the current queue of update props to our animations.
+  private _flush(onEnd?: OnEnd) {
```
aleclarson: Moved queue execution logic into the new `_flush` method.

src/animated/Controller.ts
```diff
@@ -273,20 +279,28 @@ class Controller<DS extends object = any> {
...
+      const stopToken = {}
+
       let i = 0
       let last: Promise<any>
       queue = queue.then(() =>
         to(
           // next(props)
           p => {
-            if (guid !== this.guid) return
+            if (guid !== this.guid) throw stopToken
```
aleclarson: Bail out of async scripts when (1) new updates were applied, or (2) the `stop` method was called.

src/useTrail.js
```diff
@@ -36,7 +36,7 @@ export const useTrail = (length, props) => {
           attach: attachController && (() => attachController),
         }
       }),
-    [length, updateProps.reverse]
+    [length, updateProps.config]
```
aleclarson: This `useEffect` call inside `useTrail` wasn't using `updateProps.reverse`, but it *was* using `updateProps.config`, so I fixed the deps array.

drcmda:
This pull request **fixes 2 alerts** when merging ba71d6672214a2059689252c5c1a5bfa7893aba0 into 46a357e89e8b8360cd1b581078a8c1dae043f33e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-6bf96def010a81f2c14ffc841a4327a6b33fb355)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging fbb40e49ac0fb76723d0a2248975b7f996730361 into 46a357e89e8b8360cd1b581078a8c1dae043f33e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-9ff180294757e76a7d606b91a6c80e123caa46fc)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 341b2a0628318b29c388b61ea28d0ff3b16327bc into 756ea3996529b5c92f27a36b94e70839638a4a2e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-e6c37df0b8536a81ea37b7f08b220fd53ad19474)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 70783087fe944d7ac6f37ca3e28669babaa07b7e into 756ea3996529b5c92f27a36b94e70839638a4a2e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-7522666452970909b7ce30c0b51084836b2ea0c8)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 03a28ae6079a73986576cc050630f788de55efd0 into 756ea3996529b5c92f27a36b94e70839638a4a2e - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-97ee4b0be1d463618e8b75990d00d27877dd87c5)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -190,310 +198,322 @@ class Controller<DS extends object = any> {
...
-    // We discard async calls in that case
-    const guid = ++this.guid
     const queue = (this.prevQueue = this.queue)
     this.queue = prevQueue
+    prevQueue.length = 0
 
     // Track the number of running animations.
     let runsLeft = queue.length
-    this.runCount = runsLeft
+    this.runCount += runsLeft
```
aleclarson: The `runCount` now accumulates across `_flush` calls. Since the `onRunEnd` callback will always be called, it wouldn't make sense to reset `runCount` on every flush.

src/animated/Controller.ts
```diff
@@ -190,310 +198,322 @@ class Controller<DS extends object = any> {
...
-      to = {} as any,
-      config = {},
-      reverse,
-      attach,
-      reset,
-      immediate,
-      onStart,
-    } = this.props
+  // Merge every fresh prop. Return false if no props were fresh.
+  private _diff({ timestamp, ...props }: UpdateProps<State>): boolean {
```
aleclarson: Now, the `_diff` method only updates `this.props` and `this.timestamps`, while the new `_animate` method updates `this.animations`.

src/animated/Controller.ts
```diff
@@ -190,310 +198,322 @@ class Controller<DS extends object = any> {
...
-          if (node && !isImmediate) {
-            started.push(name)
-          }
+    // Reduces input { key: value } pairs into animation objects
+    for (const key in this.merged) {
+      const state = this.animations[key as any]
+      if (!state) {
+        console.warn(
+          `Failed to animate key: "${key}"\n` +
+            `Did you forget to define "from.${key}" for an async animation?`
```
aleclarson: Added a nice warning for when `from[key]` is not defined when it should be. 🎉

drcmda:
This pull request **fixes 2 alerts** when merging 4f7e2c512790fe13a453bb5464017ef3d328117e into 370856f1d7a648cc9d805533fdf3c7577f7121e9 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-70a42b9bb919d02f357820e4ca18e58385472872)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 6afaa6bdfe61296e4f7bb069c3c4f24081faf9d7 into 370856f1d7a648cc9d805533fdf3c7577f7121e9 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-a3365ec8dbec7d52f35e4eb06c7cfa17f66d3ac5)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
drcmda: I never like onStart tbh, these prop-triggered lifecycles are all vague

src/animated/Controller.ts
drcmda: Makes sense!

src/animated/Controller.ts
aleclarson: Beyond debugging, I don't think `onStart` has much utility. That's why I changed it to be called for every `Animation` object that is about to start.

drcmda:
This pull request **fixes 2 alerts** when merging a6e3637bd958537f147b6f10a4ad6bbe57397954 into 370856f1d7a648cc9d805533fdf3c7577f7121e9 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-57214f1916ad17213fdde90f9934e9104ecf6089)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

src/animated/Controller.ts
```diff
@@ -281,129 +296,111 @@ class Controller<DS extends object = any> {
...
           reset ||
-          (hasNewGoal && currentValueDiffersFromGoal) ||
-          hasNewConfig
+          !is.equ(goalValue, entry.goalValue) ||
+          !is.equ(toConfig, entry.config)
         ) {
-          // Convert regular values into animated values, ALWAYS re-use if possible
-          if (isNumber || isString)
-            parent = interpolation =
-              entry.parent || new AnimatedValue(fromValue)
```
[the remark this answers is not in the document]
drcmda: Oh no, i see it ...

src/animated/Controller.ts
drcmda: i agree, docs are a sore spot still

drcmda:
This pull request **fixes 2 alerts** when merging 15b3107c2945f2a62e1b211f6ead4fe7c3bc8b0b into ca4f2c754d6f9c0e02f5808c0dee6ab01f4134a6 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-30055567a4b44f5f2a7368d327a38d148f80c0a7)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 8a7b3ac47043d5b717e16a43ec2675ae2cf04c28 into ca4f2c754d6f9c0e02f5808c0dee6ab01f4134a6 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-4593abc1dcfd895fb0ccf1507cbea983095e18ca)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging b9e4ff5822b60ffde97f5f6ff2941d1ed10b31d0 into ca4f2c754d6f9c0e02f5808c0dee6ab01f4134a6 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-0f9942d0071daea87d751841ce83113deb0cbe3d)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

drcmda:
This pull request **fixes 2 alerts** when merging 0ba16938ddc111999049ecc22004757954b4183b into ca4f2c754d6f9c0e02f5808c0dee6ab01f4134a6 - [view on LGTM.com](https://lgtm.com/projects/g/react-spring/react-spring/rev/pr-8d68f26348db0ac9ba0d5b7be0a0e46d469ab649)

**fixed alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused variable, import, function or class

---

PR #626, opened 2019-04-07
fix: ordering of deleted items

aocenas:
This tries to fix the ordering of deleted items after merging them to out array. First tried in https://github.com/react-spring/react-spring/pull/605 but that had some issues. This tries to be more comprehensive fix of the reconciliation logic.

This seems to work on all things I tried (but again somebody should check after me as I did not notice the last issue in the examples), but I feel it can still be flaky. When there is a single item, its sibling pointer points to itself. So if items are added and deleted one by one there is no way to know the ordering. Right now the only example with notifications work, but I think it only works because the items are added/deleted in one specific order and would again break in other cases.

So I think this is a bit of an upgrade and should solve ordering when deleting multiple items in any order, but definitely does not solve all the ordering issues. For that there will probably need to be some refactoring of the items tracking in the transition animation.

### DIFF
--- a/src/renderprops/Transition.js
+++ b/src/renderprops/Transition.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import Spring from './Spring'
 import Keyframes from './Keyframes'
 import { callProp, toArray, interpolateTo } from './shared/helpers'
+import { reconcileDeleted } from '../shared/helpers'
 
 let guid = 0
 let get = props => {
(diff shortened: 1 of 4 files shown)


drcmda:
thanks @aocenas ! @aleclarson let's merge this for v9?

aleclarson:
Sounds good! I'll review this PR sometime soon.

aleclarson:
Merged into #632 🎉 

Great work! Thanks for contributing. ❤️

Issue #621, opened 2019-04-04
useTransition "update" prop applies when order is unchanged


Andarist:
https://codesandbox.io/s/q9n0lzmoq

Basically if have a list of `[A]` and we rerender without reordering, removing or deleting items such A item gets marked as updating - thus triggering the update transition. It's not a desirable effect though.

Fixing this is kinda tricky, because it's hard to cover all cases automatically without knowing the layout constraints 🤔

drcmda:
Updates are normally all items that don't enter or leave. There's an update lifecycle that adds 100 to height, so 200something. What would you expect it to do?

Andarist:
the expected result here would be that if the list stays the same between renders no transition should happen, most notably if that list has pending transitions they shouldnt be overriden by the update transition

drcmda:
When do you expect update to kick in? The pending animation thing makes sense to me but there it gets hard for me to even say what should take precedence. The current ruleset is that updates are allowed to cut into ongoing animations, but the ongoing animations values should be merged, so we don't loose values. If enter animates color:blue, and update left:100, it should do both.

Andarist:
> When do you expect update to kick in?

Do you mean react update or spring update?

> The current ruleset is that updates are allowed to cut into ongoing animations, but the ongoing animations values should be merged, so we don't loose values.

This makes complete sense to me - if the change is detected, the merging principal is great - the animation should preserve its state and aim to land in the merged state, unless we override a particular value.

Sorry that I can't prepare a react-spring specific demo right now, I'm short on time - but this demo based on react-pose can demonstrate the issue quite nicely - https://codesandbox.io/s/l48wl5jno9 . It just has different animation config for both enter & update states and it suffers from the same issue as `useTransition` hook - it marks the item as updating when the owning component receives unrelated update (and thus render gets called and hook gets recalculated).

drcmda:
that's the general problem with using props as update trigger. for the other hooks we have 

    const [props, set, cancel] = useXXX(() => config)

that makes the config immutable and impervious to parental updates. but it isn't implemented for transition right now. would it solve the issue?

Andarist:
Maybe - more imperative API would certainly give bigger control but would lose the ease of use. Not sure what's the ideal thing to do here - I've kinda only wanted to start a discussion around this to see if you have any ideas how this could be solved.

drcmda:
[...]

PR #635, opened 2019-04-15
Fix: don't call  useTransition's update lifecycle if items are the same

Nizarius:
Following my comments on #621.

This small fix makes useTransition work as expected: 
if we don't update items props, it shouldn't call update.

Indispensable fix for functional components that have more than one prop. At the same time, the solution might not be the best one so wait your comments.


Nizarius:
Please, give me some feedback. I also want to add extraKeys as described:
https://github.com/react-spring/react-spring/issues/621#issuecomment-483324982

But I need to know that this conception is OK. At least, in my project it is silver bullet and, I suppose, many developers would like to have such an ability from the box.

aleclarson:
So basically, users would memoize the `items` argument to prevent `update` animations?

I think memoizing the `update` object might be better. WDYT?

PR #652, opened 2019-04-24
Add dependencies Array to useSpring and useSprings

Nizarius:


aleclarson:
This will be available in the next beta

aleclarson:
Note: This could use an example in https://github.com/react-spring/react-spring-examples

PR #670, opened 2019-05-15
Adapt to unified v9 api

drcmda:
- [ ] Update typings
- [ ] Fix useMemo issue (https://github.com/react-spring/react-spring/pull/670/files#r291721336)
- [ ] Support legacy API by checking `arguments.length`

### DIFF
--- a/packages/core/src/useTransition.js
+++ b/packages/core/src/useTransition.js
@@ -4,8 +4,8 @@ import { is, useForceUpdate, useOnce } from 'shared'
 import { Controller } from './Controller'
 
 /** API
- * const transitions = useTransition(items, itemKeys, { ... })
- * const [transitions, update] = useTransition(items, itemKeys, () => ({ ... }))
+ * const transitions = useTransition(items, itemKeys, props)
+ * const transitions = useTransition(items, itemKeys, () => props)
  */
 
 let guid = 0
@@ -25,6 +25,8 @@ const makeConfig = props => {
 }
 
 export function useTransition(input, keyTransform, props) {
+  // Coerce props to an object
+  props = useMemo(() => callProp(props), [])
   props = makeConfig({
     ...props,
     items: input,
--- a/packages/core/src/legacy.js
+++ b/packages/core/src/legacy.js
@@ -5,11 +5,11 @@ import { useTransition } from './useTransition'
 import { is } from 'shared'
 
 export function Spring({ children, ...props }) {
-  return children(useSpring(props))
+  return children(useSpring(props)[0])
 }
 
 export function Trail({ items, children, ...props }) {
-  const trails = useTrail(items.length, props)
+  const [trails] = useTrail(items.length, props)
   return items.map((item, index) => {
     const result = children(item, index)
     return is.fun(result) ? result(trails[index]) : result
--- a/packages/core/src/legacy.js
+++ b/packages/core/src/legacy.js
@@ -5,7 +5,7 @@ import { useTransition } from './useTransition'
 import { is } from 'shared'
 
 export function Spring({ children, ...props }) {
-  return children(useSpring(props)[0])
+  return children(...useSpring(props))
 }
 
 export function Trail({ items, children, ...props }) {
--- a/packages/core/src/useSpring.js
+++ b/packages/core/src/useSpring.js
@@ -2,12 +2,12 @@ import { useSprings } from './useSprings'
 import { is } from 'shared'
 
 /** API
- * const props = useSpring({ ... })
- * const [props, set] = useSpring(() => ({ ... }))
+ * const [props, set, cancel] = useSpring(props, [optionalDeps])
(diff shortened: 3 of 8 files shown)


src/animated/Controller.ts
```diff
@@ -320,6 +325,7 @@ export class Controller<State extends Indexable = any> {
 
   // Update the props and animations
   private _run(props: UpdateProps<State>, onEnd: OnEnd) {
+    if (globalConfig) props = { ...props, ...globalConfig }
```
aleclarson: This should be:
```js
if (globalConfig) props.config = { ...props.config, ...globalConfig }
```

src/useSpring.js
```diff
@@ -1,13 +1,12 @@
...
- * const [props, set] = useSpring(() => ({ ... }))
+ * const [props, set, cancel] = useSpring(props, [optionalDeps])
+ * const [props, set, cancel] = useSpring(() => props, [optionalDeps])
  */
 
 export const useSpring = (props, deps) => {
-  const isFn = is.fun(props)
-  const [result, set, stop] = useSprings(1, isFn ? props : [props], deps)
-  return isFn ? [result[0], set, stop] : result
+  const [result, set, stop] = useSprings(1, [callProp(props)], deps)
```
aleclarson: This calls the `props` argument on every render. Shouldn't it only be called once?

src/useTransition.js
```diff
@@ -31,6 +31,8 @@ const makeConfig = props => {
 }
 
 export function useTransition(input, keyTransform, props) {
+  // Coerce props to an object
+  props = callProp(props)
```
aleclarson: Same here. Don't want to call `props` on every render, I think.

aleclarson:
@drcmda I have a git tip for you! Do this to easily rebase your PR onto its base branch.

```sh
git pull --rebase origin v9
```

src/useTransition.js
```diff
@@ -32,7 +32,7 @@ const makeConfig = props => {
 
 export function useTransition(input, keyTransform, props) {
   // Coerce props to an object
-  props = callProp(props)
+  props = useMemo(() => callProp(props), [])
```
aleclarson: Nope, still not good enough. :)

This ends up caching `props` when it's an object, which isn't what we want.

aleclarson: Or is it? :P

drcmda: for transition im not sure tbh, it's undefined atm. i never considered it updateable, items can update, but i've never seen someone mutate the config (it also doesn't have an updater for this reason).

aleclarson: For all hooks, re-rendering should always queue an update (unless a deps array is given). Just for consistency.

drcmda: for the "update" lifecycle perhaps, yes. But the others are one-time slots. Enter only happens once, or leave, there's there that can be updated. Update is a valid concern though, so how do we want to define it? And is this something that v9 useTransition can do already?

aleclarson: The `enter` and `leave` props should still be updatable for future enters and leaves.

PS: We may want to remove the `update` prop now that the `update` function is always returned.

drcmda: Makes sense (enter/leave). The update prop is useful, it comes from d3 and has a specific meaning. It fires for all nodes that aren't either enter or leave. Like so: https://codesandbox.io/embed/26mjowzpr

aleclarson: It might be cool to allow custom phases instead of (or in addition to?) the "update" phase. The custom phase would be derived from each item. I haven't thought much about it yet, but it *could* be useful for multi-phase transitions.

src/animated/Controller.ts
```diff
@@ -325,7 +325,7 @@ export class Controller<State extends Indexable = any> {
 
   // Update the props and animations
   private _run(props: UpdateProps<State>, onEnd: OnEnd) {
-    if (globalConfig) props = { ...props, ...globalConfig }
+    if (globalConfig) props.config = { ...props.config, ...globalConfig }
```
aleclarson: This needs to account for `props.config` being a function, so we'll need to move this somewhere better.

drcmda: right, so _run isn't the proper place. how about diff, where it unpacks config? or does it affect multiple places?

aleclarson: I made it so `globalConfig` is merged in `_animate`, which means it doesn't get processed by `_diff`, but I think I'm okay with that.

packages/core/src/useTransition.js
```diff
@@ -25,6 +25,8 @@ const makeConfig = props => {
 }
 
 export function useTransition(input, keyTransform, props) {
+  // Coerce props to an object
+  props = useMemo(() => callProp(props), [])
```
aleclarson: Note: This still needs fixing. 👀

aleclarson:
Closed in favor of #808

PR #687, opened 2019-06-01
chore: convert to monorepo

### LINKED ISSUE #677: Convert to monorepo
There are a few reasons why switching to a monorepo is a good idea.

- The `@types/react-native` package conflicts with `"lib": ["dom"]`

- The target for React Native doesn't need to be bundled, because Metro handles compilation and caching for you. By not bundling, source maps are properly supported as well.

- Only the `renderprops` bundle uses the `prop-types` package, so we shouldn't need to install it for the other bundles.

- Cheaper installs by not aggregating all bundles in the same package.

- The `@react-spring/core` package would be able to serve the same purpose as the `cookbook` bundle.

- With each target having its own package, we can ensure the `peerDependencies` are properly defined. Same for `devDependencies`.

- Once we add "native frameloop" support to the React Native target, it would be better to only ship that code to React Native users.

And possibly some other reasons I'm not thinking of?

aleclarson:
Closes #677

This explodes `react-spring` into the following packages:

- `@react-spring/core` - Hooks, renderprops, `FrameLoop` class, and `Controller` class
- `@react-spring/animated` - The fork of [animated](https://github.com/animatedjs/animated)
- `@react-spring/web`
- `@react-spring/native` - **Note:** This isn't transpiled before it's published.
- `@react-spring/three`
- `@react-spring/konva`
- `@react-spring/zdog`
- `@react-spring/shared` - Globals and shared modules
- `react-spring` - Mostly the same as before, except the `/renderprops` and `/cookbook` entry points are gone

⚠ **TypeScript must be 3.5.0+ from now on!**

Read the `CONTRIBUTING.md` file to learn how to clone the monorepo.

You can run `yarn build` in the root or in a specific package. <s>Currently, `yarn build` runs on every package every time, even if only one package changed.</s> And there's not yet a watch mode that includes all packages, but you can run `yarn build --watch` in a specific package.

Run `lerna version && yarn release` to publish every package with a new version.

As of now, every package installs `tslib` and `@babel/runtime` (even if never imported), because the `yarn prepare` command cannot detect the actual dependencies. We could make a Rollup plugin that detects when one of these packages is imported, and it could then add to the `dependencies` field in `dist/package.json`.


brunolemos:
> @react-spring/native - Note: This isn't transpiled before it's published.

Was it before?
For react-native-web ideally it should be transpiled. There may be a way to have both, transpiled and non-transpiler, but I don't have experience with this.

aleclarson:
@brunolemos You don't use Metro with react-native-web?

brunolemos:
@aleclarson no, normal web stuff, like create-react-app.

aleclarson:
@brunolemos Is that what react-native-web recommends?

brunolemos:
@aleclarson yes react-native-web uses webpack on their [official example](https://github.com/necolas/react-native-web/tree/master/packages/examples) and create-react-app on their [issue reproduction quickstart](https://codesandbox.io/s/6lx6ql1w5r).

It's common for react-native packages to not be transpiled. When they have some unsupported syntax we have to compile them by editing webpack. It's not a huge deal if it's not transpiled, but makes life easier. If react-spring considers react-native-web an officially supported target then it might want to have a transpiled version.

aleclarson:
> It's common for react-native packages to not be transpiled. When they have some unsupported syntax we have to compile them by editing webpack.

What are the required steps to support an uncompiled package with Webpack or create-react-app? I'm thinking it will be easier to include a guide, instead of having two versions of the RN target.

brunolemos:
> What are the required steps to support an uncompiled package with Webpack or create-react-app?

Basically the user has to add the missing babel plugins and point to the node_modules/package folder to include it in the compilation. For create-react-app, to avoid ejecting, there are tools like https://github.com/arackaf/customize-cra. It has helpers like `addBabelPlugin` and `addWebpackResolve`. 

_Not a webpack pro so I'm not the best person to write detailed steps._

PR #693, opened 2019-06-05
Extend HooksBaseProps from SpringProps

gscottolson:
Currently, the typings for useTransition cause an error when declaring `onRest` and `onFrame`. This seems to be caused by a chain dependencies that result in:

```typescript
export interface HooksBaseProps extends Pick<SpringBaseProps, Exclude<keyof SpringBaseProps, 'config'>> {
```

SpringBaseProps looks like:
```typescript
export interface SpringBaseProps {
  config?: SpringConfig | ((key: string) => SpringConfig)
  native?: boolean
  reset?: boolean
  immediate?: boolean | ((key: string) => boolean)
  delay?: number
  reverse?: boolean
  onStart?(): void
}
```

The SpringProps interface declares the missing props and extends SpringBaseProps, so I think the solution is for HooksBaseProps to extend SpringProps instead of SpringBaseProps directly.

### DIFF
--- a/types/web.d.ts
+++ b/types/web.d.ts
@@ -2,6 +2,7 @@ import { CSSProperties, RefObject } from 'react'
 import {
   SpringConfig,
   SpringBaseProps,
+  SpringProps,
   TransitionKeyProps,
   State,
 } from './renderprops-universal'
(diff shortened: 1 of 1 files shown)


aleclarson:
Thanks for contributing! But it's unlikely that this will be merged, because Paul doesn't know TypeScript and I'm focusing solely on v9. Sorry!

PR #716, opened 2019-06-25
Check `onRest` to be a function

danielhusar:
Might `fix` https://github.com/react-spring/react-spring/issues/703


aleclarson:
Just pushed a fix. Sorry for the wait! 👍

PR #722, opened 2019-06-27
Improve TypeScript setter definition for useSprings

barry4dev:
The change here fixes the TS setter call for the useSprings callback - which requires an int -> new useSprings props.

Great job with react-spring - it's superb!

### DIFF
--- a/types/web.d.ts
+++ b/types/web.d.ts
@@ -71,11 +71,13 @@ type InferFrom<T extends object> = T extends { to: infer TTo }
 //  but with a delayed evaluation that still allows A to be inferrable
 type Merge<A, B> = { [K in keyof A]: K extends keyof B ? B[K] : A[K] } & B
 
-export type SetUpdateFn<DS extends object> = (ds: Partial<UseSpringProps<DS>>) => void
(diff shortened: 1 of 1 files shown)


aleclarson:
Types for v8 aren't being maintained. I recommend using v9. 

Thanks for taking time to help out, though 👍

barry4dev:
Hi @aleclarson thanks for the quick reply, I went ahead and tried v9.  My code works equally well there (great!), but the Typescript still seems to be a problem (either that or I'm using useSprings's setter wrong).

Please do check out https://github.com/react-spring/react-spring/issues/723 when you can.  Thanks

EECOLOR:
> Types for v8 aren't being maintained. I recommend using v9.

That does not make any sense. v9 is not stable yet right? I would expect bugs to be fixed in the stable version. Maybe this comes over a bit harsh (sorry if it does), but I really do not understand.

What if React had a bug and stated: it will not be fixed in the current stable version, please use the unstable version without the bug. And if we would find another bug in the unstable version the reply probably would be: yeah, this is an unstable version, do not use it in production yet.

PR #726, opened 2019-06-28
TypeScript useSprings improvements (and web package build fix)

barry4dev:
Followup from https://github.com/react-spring/react-spring/issues/723 after testing
(please make a beta so I can fully test, but this builds fine - the annoying react hook packages linkages is causing me full testing problems)
- SpringUpdateFn (the non-callback scenario) is used in multiple places and breaks the build elsewhere, so I created a new SpringsUpdateFn for this scenario
- made the spring argument optional in the callback
- the link/unlink package process was more difficult since some packages don't have dist folders
- fixed a web package build problem ([!] Error: Character is out of bounds) caused by 
```
export type JSXElements = keyof JSX.IntrinsicElements		
```
 (unrelated to my issue, but just helping out :) ) -- if someone was really wanting something like JSXElements, it's likely they shouldn't pull this from this package.


package.json
```diff
@@ -40,7 +40,9 @@
     "release": "lerna publish",
     "test": "jest",
     "test:cov": "jest --coverage",
-    "test:ts": "lerna exec --concurrency=1 --no-bail -- 'pwd && tsc -p . --noEmit'"
+    "test:ts": "lerna exec --concurrency=1 --no-bail -- 'pwd && tsc -p . --noEmit'",
+    "link-packages": "lerna exec -- '[[ -d \"./dist\" ]] && (cd dist && yarn link) || yarn link'",
+    "unlink-packages": "lerna exec -- '[[ -d \"./dist\" ]] && (cd dist && yarn unlink) || yarn unlink'"
```
aleclarson: I don't think these scripts are necessary. You only link the packages once, and it's not hard to unlink them manually. Also, the directory check is unnecessary since `yarn` triggers `yarn build` (which creates the `dist` folders).

packages/core/src/types/spring.ts
```diff
@@ -55,12 +55,22 @@ export type SpringUpdate<T extends object = {}> = Partial<T> &
...
+ * The `T` parameter should only contain animated props.
+ */
+export interface SpringsUpdateFn<T extends object = {}> {
   /** Update the props of a spring */
   (
     props:
       | SpringUpdate<T>
       | SpringUpdate<T>[]
-      | ((index: number, spring: Controller<T>) => SpringUpdate<T>)
+      | ((index: number, spring?: Controller<T>) => SpringUpdate<T>)
```
aleclarson: This is wrong, because the controller is always passed.

packages/core/src/types/spring.ts
```diff
@@ -55,12 +55,22 @@ export type SpringUpdate<T extends object = {}> = Partial<T> &
...
+  /** Update the props of a spring */
+  (props: SpringUpdate<T>): void
+}
+
+/**
+ * Imperative API for the useSprings hook, allowing to update individually or over the array
+ *
+ * The `T` parameter should only contain animated props.
+ */
+export interface SpringsUpdateFn<T extends object = {}> {
```
aleclarson: :+1:

targets/web/src/elements.ts
```diff
@@ -1,5 +1,4 @@
-export type JSXElements = keyof JSX.IntrinsicElements
-export const elements: JSXElements[] = [
+export const elements: Array<keyof JSX.IntrinsicElements> = [
```
aleclarson: This seems like an unnecessary change. Let's keep PRs focused.

package.json
aleclarson: > the link/unlink package process was more difficult since some packages don't have dist folders

Sorry, I just noticed what you said. I think using `--no-bail` will ignore the lack of a `dist` folder in `packages/envinfo`. You could also use `--ignore "*/envinfo"`. I'll update the contributor guide.

targets/web/src/elements.ts
aleclarson: > fixed a web package build problem ([!] Error: Character is out of bounds)

Were you seeing this error when run `yarn build` or something?

> if someone was really wanting something like JSXElements, it's likely they shouldn't pull this from this package.

The `JSXElements` type isn't re-exported, so I don't think that's an issue.

aleclarson:
Merged into `v9` here: 51d9fb8aad67da1182a73798fd658fa015109a97

aleclarson:
Btw, thanks for being the first outside contributor to v9! 🎉

targets/web/src/elements.ts
aleclarson: Looks like refreshing `yarn.lock` fixed this 🎉

packages/core/src/types/spring.ts
barry4dev: ah, my setter call didn't need a second arg, but I guess you're defaulting it?

barry4dev:
Sweet - thanks! 🎉

barry4dev:
Re the wording spec, would be good to put that in the contributing guide, but I'll do ti going forward for sure. It would be good to put that in the contributing as a reminder.  I'm a bit new to open source contributions.  (I can rebase to reword them if you want?)

packages/core/src/types/spring.ts
aleclarson: Sorry, not sure what you mean. The `?` type operator only affects the caller. The callback is allowed to omit any trailing arguments it doesn't need.

```ts
const [props, set] = useSprings(i => { ... })

// Both of these calls are valid in TypeScript's eyes
set(i => { ... })
set((i, spring) => { ... })
```

barry4dev: ah of course - thanks!

PR #728, opened 2019-06-30
docs: add section for react hooks help for resolving react for hooks

barry4dev:
Removed section on exclusive-deps since it doesn't connect monorepo dependencies (whereas yarn link does)


aleclarson:
I reworded it some. Thanks! :+1:

PR #750, opened 2019-07-10
feat: rework useTransition

aleclarson:
This PR is a rewrite of `useTransition` from scratch in hopes of simplifying both the API and the internals. Here's an example of the new API:
```tsx
import {animated, useTransition} from 'react-spring'
const MyComponent = ({ items, ...props }) => {
  const transition = useTransition(items, props)
  return transition((values, item) => (
    <animated.div style={values}>{item}</animated.div>
  ))
}
```
Once #670 is merged, you can pass a deps array and you'll get the `set` and `cancel` functions returned to you:
```ts
const [transition, set, cancel] = useTransition(items, props, [...deps])
```

## Notable changes

- No more `keys` argument, and the `key` prop is handled for you
- Returns a `transition` function instead of a `transitions` array
- Now safe to use in concurrent mode
- Unique mode is the new default (and it's not configurable)
- Lazy mode is the new default
- Event props cannot be passed directly anymore, but the phase props still support them
- The `reset` prop now uses the `initial` prop if it exists

### New `expires` prop

For controlling when deleted items are dismounted after their leave animation finishes. Set to `0` for immediate dismount. By default, they avoid dismounting until all transitions are resting. Any value above `0` is treated as "milliseconds to wait before dismounting is forced".


aleclarson:
## Notes

- We need a demo that uses `z-index` to overcome the lack of DOM order guarantee.
- Currently, "expired" transitions are dismounted whenever the component re-renders, even if other controllers are still animating. Is this good or bad? 🤔

PR #753, opened 2019-07-10
Types: Stop Animation Function in Tuple

avin-kavish:
The return tuples of the hooks did not have the 3rd value declared. Someone had left the following comment,
> there's a third value in the tuple but it's not public API (?)

Since the documentation demonstrates how to use the function returned to stop the animation, it's safe to say it's part of the public API, so I added it to the type definitions.

I considered placing `() => void` everywhere but decided against it and used the alias `StopAnimationFn` to make the purpose of the function clear.

### DIFF
--- a/types/web.d.ts
+++ b/types/web.d.ts
@@ -124,15 +124,15 @@ export type UseSpringProps<DS extends object> = Merge<
 
 type OverwriteKeys<A, B> = { [K in keyof A]: K extends keyof B ? B[K] : A[K] };
 
-// there's a third value in the tuple but it's not public API (?)
+type StopAnimationFn = () => void
+
 export function useSpring<DS extends object>(
   values: UseSpringProps<Merge<DS, CSSProperties>>
 ): AnimatedValue<ForwardedProps<OverwriteKeys<DS, CSSProperties>>>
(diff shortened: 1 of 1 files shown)


aleclarson:
Sorry, we're not maintaining the typings for v8 anymore. Please use v9 beta if you can.

```
npm install react-spring@next
```

avin-kavish:
How different is v9? Are the tuples gone?

aleclarson:
Test your animations locally with v9 and report any issues in #642. Make sure to pin the version instead of using caret ranges, and you'll be fine. [Intercom](https://www.intercom.com) is using the beta in their flagship product already.

avin-kavish:
I'll do that but, 

1. this sort of mishap wouldn't happen if master were the bleeding edge. Or if v9 were the default branch for the repo and if there were some more information on the state of the project. Currently, the readme is only "how to", and "who uses"

2. Is v8 considered beyond repair? I noticed unpredictable behaviour with transitions and broken types. I'm not sure if you looked at the changes, but this is a pretty simple change that does not attempt to fix any of the broken types but just adds the stop function to the returned tuple. There are no drastic changes.

PR #760, opened 2019-07-19
FIX: Add onRest type to use transition props

alexknipfer:


AquiGorka:
@aleclarson I am getting a typescript error with `useTransition` + `onRest` (`react-spring@8.027`):

```
 No overload matches this call.
  Overload 1 of 2, '(items: ISection | readonly ISection[] | null | undefined, keys: string | number | readonly ReactText[] | ((item: ISection) => ReactText) | null, values: Merge<CSSProperties, UseTransitionProps<ISection, CSSProperties>>): UseTransitionResult<...>[]', gave the following error.
    Argument of type '{ config: { mass: number; tension: number; friction: number; }; enter: { transform: string; }; from: { transform: string; }; initial: null; leave: { pointerEvents: "none"; position: "absolute"; transform: string; }; onRest: () => void; }' is not assignable to parameter of type 'Merge<CSSProperties, UseTransitionProps<ISection, CSSProperties>>'.
      Object literal may only specify known properties, and 'onRest' does not exist in type 'Merge<CSSProperties, UseTransitionProps<ISection, CSSProperties>>'.
```

Do you think it is related to this issue?

PR #632, opened 2019-04-10
v9.0.0

aleclarson:
Follow-up to #615

- [x] rework the Controller class (a lot)
- [x] rework the TypeScript types (with tests!)
- [x] rewrite the Render Props API on top of hooks (#561)
- [x] merge #626
- [x] look for issues fixed by this PR

### [Changelog](https://www.react-spring.io/log)

Closes #335
Closes #362
Closes #383
Closes #432
Closes #436
Closes #461
Closes #535
Closes #559
Closes #571
Closes #576
Closes #591
Closes #594
Closes #613
Closes #623
Closes #624
Closes #629
Closes #633
Closes #634
Closes #637
Closes #641

PR #761, opened 2019-07-22
FIX: types ParallaxLayer and Parallax props to real implementation and opportunities

RUVATA:
Adding correct type definition to props signature of real implementation of Parallax and ParallaxLayer exported classes. Support for props translated to native elements 'as is'

### DIFF
--- a/types/renderprops-addons.d.ts
+++ b/types/renderprops-addons.d.ts
@@ -1,7 +1,7 @@
 import { Ref, PureComponent } from 'react'
 import { SpringConfig } from './renderprops-universal'
 
-interface ParallaxProps {
+interface ParallaxProps extends React.HTMLAttributes<HTMLElement> {
   pages: number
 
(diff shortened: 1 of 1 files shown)


RUVATA:
In fact the ParallaxLayer and Parallax (wrapper) classes signature should implements HTMLDivElement and abstract HTMLElement.
Placed to them properties are provided to native element 'as is', realization talks thats it is intention of developer.
Another argument that the official example of usage of Parallax, actively used placement props as 'style', 'onClick' handler and other. 

This is fix in type declarations - allows to use props translated to native elements at rules of type guard, without TypeScript (TS2322) errors, because in fact user can setup to Parallax & ParallaxLayer any property compatible with HTMLDivElement and abstract HTMLElement signatures, it need to be supported in type definition.

RUVATA:
the problem will not be relevant in #632

PR #788, opened 2019-08-20
feat: add "config.step" prop

aleclarson:
This rounds the animated value to the nearest step.

For example, passing `10` as the step would round to the nearest multiple of 10.

Likewise, passing `0.5` would round to the nearest multiple of 0.5.

The most common use case will probably be rounding based on pixel density.


aleclarson:
Merged into #808

PR #790, opened 2019-08-26
Better inheritance of config properties and callbacks

dbismut:
## Feature proposal

Better inheritance of controller config properties following the logic below:

- Properties defined in the initial `useSpring` call should be extended on subsequent  `set` calls. 
- Properties defined in a `set` call should override initial properties for that call only (ie a second `set` call shouldn’t inherit properties defined in the first one).
- In async runs, `next` calls should inherit properties from the outer `set`, with the exception of `onAnimated` which should be called once at the beginning of the animation, and `onRest` which should be called once at the end of the animation. However, defining `onFrame` in set would apply to all inner `next` runs.
- Similarly, consecutive `next` runs shouldn’t inherit properties from each other.

### Example

```jsx
console.log('executing initial spring call')

const [props, set, stop] = useSpring(() => ({
  y: 0,
  x: 0,
  onRest: () => console.log('onRest initial'),
}))

React.useEffect(() => {
  console.log('executing set call')

  set({
    config: config.wobbly,

    // should execute only once
    onAnimate: v => console.log('[set] onAnimate'),

    // should execute on each frame
    onFrame: v => console.log('[set] onFrame'),

    // should execute only once, will override initial onRest
    onRest: v => console.log('[set] onRest'),
    to: async next => {
      console.log('executing first next call')

      await next({
        y: 200,

        // config will be stiff when animating `y`
        config: config.stiff,

        // should override onFrame defined in above
        onFrame: v => console.log('[next] onFrame'),
      })

      console.log('executing second next call')
      // should execute onFrame function defined in the `set` call
      // should apply config.wobbly when animating `x`
      await next({ x: 300 })
    },
  })
})
```

#### Output

The console should output the following logs:

options

### Disclaimer

This is a very naive implementation and should probably be reworked.

### DIFF
--- a/packages/core/src/Controller.ts
+++ b/packages/core/src/Controller.ts
@@ -45,6 +45,7 @@ type PendingProps<State extends object> = Merge<
     parent?: Controller | null
     attach?: (ctrl: Controller) => Controller | null
     timestamp?: number
+    async?: boolean
   }
 >
 
@@ -56,6 +57,29 @@ type CachedProps<State extends object> = Merge<
     asyncTo?: ReadonlyArray<SpringUpdate<State>> | SpringAsyncFn<State>
   }
 >
+
+type Options<State extends object> = {
+  initial: CachedProps<State>
+  set: CachedProps<State>
+  next: CachedProps<State>
+}
+
+type OptionKey =
+  | 'onAnimate'
+  | 'onStart'
+  | 'onFrame'
+  | 'onRest'
+  | 'immediate'
+  | 'config'
+
+const options: OptionKey[] = [
+  'onAnimate',
+  'onStart',
+  'onFrame',
+  'onRest',
+  'immediate',
+  'config',
+]
 
 // Default easing
 const linear = (t: number) => t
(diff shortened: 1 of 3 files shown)


aleclarson:
When an inherited `config` prop exists, do you think it should be merged with an update-specific `config` prop? Maybe it depends on whether the inherited `config` is a function or object? At the moment, I'm going with "only use the inherited `config` when an update has no `config` of its own defined".

dbismut:
What’s a bit tricky is that I don’t think `onAnimate` or `onRest` should trigger on each async calls if you’ve just set them up in the update function. So unless you’ve actually specified `onAnimate` or `onRest` in `next` config, I believe they should trigger at the beginning of the animation and at the end if they’re specified either in the initial props or in the `update` config.

Does that make sense?

aleclarson:
Sounds good! You didn't answer my question though. lol

dbismut:
Haha I was afraid of me not understanding your question and that is the case it seems!

drcmda:
> In the example above, should the config from the update call be merged with the default config (eg: { mass: 2, friction: 100 }) or just override it (eg: { mass: 1, friction: 100 })?

i like merging, but then also got in trouble often because of it. its easier at first glance but can become an obstacle. override is more explicit i think. this forces the user to create self contained presets.

aleclarson:
Sounds good! **Note:** You can always pass `default: true` to mutate the default props with an `update` call. 🎉

aleclarson:
One more question! Any objections to updates only affecting the keys that are defined in the `from` and/or `to` props? For example:

```ts
update({ x: 100 }) // use default "config"
update({ z: 100, config: { mass: 2 } }) // this only affects "z" not "x"
```

PR #792, opened 2019-08-27
Using CASpringAnimation inspired algorithm for spring

dbismut:
## Description

Most of it is already explained here: https://github.com/skevy/wobble#background

I'm not exactly sure about the performance improvements, I've tried to run some benchmarking tests with [Benchmark](https://github.com/bestiejs/benchmark.js) and the new algorithm seemed to get slightly more ops/sec but I don't think this is reliable enough to conclude anything.

I do believe initial velocity is better handled using this algorithm though. 

### Minor improvements
This PR also adds velocity for both decay and easing animations.

### DIFF
--- a/packages/animated/src/AnimatedValue.ts
+++ b/packages/animated/src/AnimatedValue.ts
@@ -16,6 +16,7 @@ export class AnimatedValue<T = unknown> extends Animated
   lastVelocity?: number
   startTime?: number
   lastTime?: number
+  elapsedTime: number = 0
   done = false
 
   constructor(value: T) {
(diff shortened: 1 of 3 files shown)


dbismut:
> These velocity calculations can be moved to another PR if we decide to pass on the wobble algorithm.

Indeed. I'm working on another PR that also implements @mtiller bounce idea for clamp:

> Let's say you go past the min or max value. What you can do is simulate the behavior of a bouncing ball (shameless plug). What that means is that whenever you hit the limit, you set the position to the limit value (min or max, depending on which way you are going) but you reverse the value of the velocity and multiple it by what is called the "coefficient of restitution". This will trigger a "bounce" away from the mechanical limit. In this way, you never get a position value that is outside the domain you want.

> But what if you want the "clamp" behavior? Easy...make the coefficient of restitution zero. Then you get a perfectly plastic collision which means the object just stops where it is. As long as your to value isn't outside your domain, you should be just fine.

This is the result for a coefficient of restitution equal to 2:

ScreenFlow

> We should probably do some "load testing" for perf comparison

I've actually run some tests on jsperf: https://jsperf.com/react-spring/5

To prevent `requestAnimationFrame` from throttling ops/second, I've replaced it with `setTimeout(fn, 0)` to measure the raw algorithm performance.

The CASpringAnimation algorithm is a clear winner over, but I'm pretty sure there's an issue with the tests themselves: when I tested the react-spring-examples against CASpringAnimation, the svg bubble morph example was distinctively slower on my Macbook Pro.

dbismut:
Closing in favor of #797.

PR #797, opened 2019-08-29
Minimize steps necessary to accomplish semi implicit Euler interpolation 

dbismut:
### Performance
This PR tries to implement indications given in this [comment](https://github.com/react-spring/react-spring/issues/678#issuecomment-498066662), with `dt` being related to the natural frequency (here `300/naturalFrequency`) rather than always being `1ms`, thus reducing the number of steps necessary to compute each frame.

**Observation**
In my observations, I've noticed slightly different behaviors of the springs when applying this change (usually the springs feel slightly faster). I might be doing something wrong.

**Possible improvements**
Here natural frequency is computed on each frame, although being a constant. It could be calculated only when tension or mass change.

### Bounce behavior on clamp
Again, as @mtiller suggested, clamp could behave like a bouncing ball. This PR allows `config.clamp` to be a number, treated as the coefficient of restitution. Below is an animation of `config.clamp = 2`.

ScreenFlow

This PR also computes `lastVelocity` and `lastTime` for easing and decay.


packages/core/src/FrameLoop.ts
```diff
@@ -159,50 +172,53 @@ export class FrameLoop {
...
         if (finished) to = position
       }
       // Spring easing
       else {
-        let lastTime = animated.lastTime !== void 0 ? animated.lastTime : time
-        if (animated.lastVelocity !== void 0) {
-          velocity = animated.lastVelocity
-        }
+        velocity =
+          animated.lastVelocity !== void 0 ? animated.lastVelocity : v0 * 1000
```
aleclarson: Why does `v0` need `* 1000`?

packages/core/src/FrameLoop.ts
```diff
@@ -159,50 +172,53 @@ export class FrameLoop {
...
+
+        const freq = Math.sqrt(config.tension! / config.mass!) * (2 * Math.PI)
```
aleclarson: Why is `* (2 * Math.PI)` needed?

aleclarson:
Can you open a PR on [react-spring-examples](https://github.com/react-spring/react-spring-examples) for the example in that gif?

dbismut:
> Run `git pull --rebase` to sync up with v9 again, please.
Here's the steps I made to synced the forked branch:
```
git fetch upstream  
git checkout v9
git merge upstream/v9
```

And then I rebased this branch onto v9 via Github desktop. Again apologies if this is not the way to go, but `git pull --rebase` didn't do anything.

dbismut:
> Can you open a PR on [react-spring-examples](https://github.com/react-spring/react-spring-examples) for the example in that gif?

Done! https://github.com/react-spring/react-spring-examples/pull/21

packages/core/src/FrameLoop.ts
dbismut: This is an error from my end, but I realize velocity for spring is inconsistent between decay, easing and spring. I suspect spring velocity is expressed `px/seconds` while decay and easing is expressed in `px/milliseconds`, so there must be a need for harmonization there. In `react-use-gesture` velocity is expressed in milliseconds, I'll change it to whatever should be `react-spring` default.

@drcmda any advice?

packages/core/src/FrameLoop.ts
```diff
@@ -159,50 +172,53 @@ export class FrameLoop {
...
+        const dt = 300 / freq
```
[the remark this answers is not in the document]
dbismut: No, actually the higher the number, the less steps you'll make and the more performance you'll get, but with the risk of reducing the precision. @mtiller (sorry for all the notification he's getting) suggested to start with `500`, `300` *felt* more natural and close to `dt = 1`. My observation is that the larger the step, the higher the velocity is computed at each step. I'm trying to investigate on this.

aleclarson:
> but `git pull --rebase` didn't do anything.

Whoops, I meant `git pull --rebase v9` :P

packages/core/src/FrameLoop.ts
dbismut: I've just pushed code to express velocity in `ms`.

packages/core/src/FrameLoop.ts
dbismut: Ok, so I asked help around to sort this out and just pushed new code:

```js
const w0 = Math.sqrt(config.tension! / config.mass!)
const dt = 100 / w0
const numSteps = Math.ceil(deltaTime / dt)

for (let n = 0; n < numSteps; ++n) {
  const springForce = -config.tension! * (position - to)
  const dampingForce = -config.friction! * (velocity * 1000)
  const acceleration = (springForce + dampingForce) / config.mass!
  velocity = velocity + (acceleration * dt) / (1000 * 1000) // expressed in ms
  position = position + velocity * dt
}
```

**Disclaimer: the following is only my understanding**

I think I was confused by the term natural frequency, which isn't what `w0` is. The correct term for `w0` is angular frequency.

The stability criterion from the [article, equation 23 page 23](https://hplgit.github.io/num-methods-for-PDEs/doc/pub/vib/pdf/vib-4print-A4-2up.pdf) @mtiller was referring to is valid only for undamped harmonic oscillator. I suspect that the right formula for damped harmonic oscillator is actually `dt < Math.sqrt(2) / w0` instead of `dt < 2 / w0`.

The stability criterion makes sure your oscillator is stable (meaning it will get to its destination eventually), but there will be a margin of error, and the closer you get to the stability limit, the more the spring will behave erratically compared to the analytical solution.

Here, I've set `dt = 0.1 / w` (actually `dt = 100/w` as these are `ms`), and it seems to behave pretty close to when `dt` was `1ms`. My observation is that we move from **10-20 steps** to just **2-3 steps** in average for the following config: `config: { tension: 120, friction: 12, mass: 1 }`.

aleclarson:
FYI: I force pushed to sync with `v9` branch.

I'll test out this PR in ~24 hours. 👍

dbismut:
@aleclarson I’m currently working on an example that will let you change the steps, alternate between analytical and Euler algorithm, plot the results on a graph, and make performance analysis. Obviously this would use a modified version of this fork, but I’d love to publish this somewhere online for you to have a look.

aleclarson:
Nice! I'll wait until then to try this PR out. Just setup a webpack repo with the example that uses your fork. That should be good enough for now. 👍

dbismut:
@aleclarson Just finished, not sure how easy this is to set up for you but:
- here is the react-spring fork: https://github.com/dbismut/react-spring/tree/feat/performance
- here is the react-spring-examples fork: https://github.com/dbismut/react-use-gesture-examples/tree/v9-performance

It should look like this:
perf

I've made some tests and I'm not sure about the conclusions just yet. It looks like increasing `dt` has some severe precision impact on low tension springs. Performance is also pretty erratic to measure, that might need some improvements.

tim-soft:
@dbismut wow this is super interesting, I setup the v9-performance branch of your example but am getting some errors, although maybe I missed something

[x.elapsedTime is always undefined](https://github.com/dbismut/react-use-gesture-examples/blob/v9-performance/demos/hooks/performance/index.js#L32)

[Chart data](https://github.com/dbismut/react-use-gesture-examples/blob/v9-performance/demos/hooks/performance/Chart.js#L19) winds up looking like this for me
```js
{
  cycles: 124
  label: "euler (100/ω)"
  max: undefined
  mean: NaN
  sum1000: undefined
}
```

dbismut:
Hey @tim-soft are you *also* using the performance branch from react-spring in my repo? In any case the performance evaluation is messy atm.

dbismut:
Here's an online version (probably unoptimized): https://spring-perf.surge.sh/

aleclarson:
> Here's an online version (probably unoptimized): [spring-perf.surge.sh](https://spring-perf.surge.sh/)

Lookin good! Interesting how the "euler" method ends earlier. Also, the overshooting last longer with the "analytical" method, perhaps long enough that this should be considered a breaking change. Seems that people will need to tune their configs.

dbismut:
@aleclarson At the moment there’s nothing from the analytical method that is a certain improvement over semi implicit Euler with a 1ms step. This PR does improve two things for sure:

- adds velocity to duration-based and decay config
- sets velocity in milliseconds for springs

Minimizing steps with `dt = 100/w0` is not conclusive at the moment for a low tensions as you can see from the screenshot I shared earlier.

packages/core/src/FrameLoop.ts
```diff
@@ -125,86 +125,100 @@ export class FrameLoop {
...
+          velocity = velocity + acceleration * dt // pt/ms
+          position = position + velocity * dt
         }
 
-        animated.lastTime = time
-        animated.lastVelocity = velocity
-
         // Conditions for stopping the spring animation
-        const isOvershooting =
-          config.clamp && config.tension !== 0
```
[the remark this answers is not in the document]
dbismut: Oh the check is there to make sure this is a spring-based config, right so that clamp doesn’t apply to duration/decay based animations?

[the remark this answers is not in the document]
aleclarson: No, in that case, `tension` would be undefined. And this check is performed after it's already known to be a spring animation. Maybe do a test with `tension` of 0 and see if anything weird happens?

aleclarson:
Side note: I've had issues with `config.precision` in the past, where animations would take a long time to rest because I needed a low (`0.001`) displacement threshold. They would oscillate for a long time around the end value, but no visual changes would happen. So maybe the velocity precision should be a constant, and keep `config.precision` for displacement only. The reasoning here is that resting velocity is the same for all animations, but this assumption needs to be tested by comparing the resting velocity of a `0 -> 1` animation with the resting velocity of a `0 -> 100` animation, for example.

dbismut:
About your side note, the velocity would be `100` times inferior for a displacement from `0 -> 100` compared to `0 -> 1`, so I don’t think precision should be a constant for velocity. However, because the velocity used to be expressed per seconds, the precision was too fine by a factor of `*1000`, that might explain the issues you’ve had. It would be great if you could test your usecase again using this PR if you can!

dbismut:
Btw I’m thinking default precision *could* be a fraction of the displacement magnitude.

packages/core/src/FrameLoop.ts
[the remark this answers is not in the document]
dbismut: `tension === 0` has the same visual effect as `immediate: true`. Putting it back in.

dbismut:
> Btw I’m thinking default precision _could_ be a fraction of the displacement magnitude.

Not sure how to do this properly @aleclarson 😅 I think it should be something like `Math.abs(to - from) / 1000` in my opinion.

packages/animated/src/AnimatedValue.ts
```diff
@@ -60,8 +61,9 @@ export class AnimatedValue<T = unknown> extends Animated
       this.startPosition = this.value
       this.lastPosition = this.value
       this.lastVelocity = isActive ? this.lastVelocity : undefined
-      this.lastTime = isActive ? this.lastTime : undefined
       this.startTime = G.now()
+      this.lastTime = isActive ? this.lastTime : this.startTime
+      this.elapsedTime = isActive ? this.elapsedTime : 0
```
aleclarson: Can `decay` and `easing` animations actually continue seamlessly from a previous update? I doubt it, but I haven't tried. If not, let me know and I'll push a local commit I have that changes this.

dbismut: I haven’t touched this so I suspect they don’t. `elapsedTime` is just a timer that accounts for lost frames for `decay` and `easing` :)

dbismut: I’m on my phone but the commit looks perfect, I guess `progress` only makes sense for `easing`?

dbismut:
@aleclarson some of our commits conflicted as I was trying to use `raf` step and therefore use a common step for all animated values in the frame loop. I've asked @drcmda if it was ok but he told me that `react-three-fiber calls` updates manually and that might interfere with this idea. I'm waiting for him to get back to me.

As far as I can see, it seems to work properly.

packages/core/src/FrameLoop.ts
```diff
@@ -62,11 +64,18 @@ export class FrameLoop {
 
     this.update =
       (update && update.bind(this)) ||
-      (() => {
+      ((time?: number) => {
         if (this.idle) {
           return false
         }
 
+        time = time !== void 0 ? time : Date.now()
```
aleclarson: Use `performance.now` instead, since that's what rAF gives us.

packages/core/src/FrameLoop.ts
```diff
@@ -111,9 +122,11 @@ export class FrameLoop {
   }
 
   /** Advance an animation forward one frame. */
-  advance(config: ActiveAnimation, changes: FrameUpdate[2]): boolean {
-    const time = G.now()
-
+  advance(
+    step: number,
```
aleclarson: Let's rename `step` to `dt` to be more clear

packages/core/src/FrameLoop.ts
```diff
@@ -187,7 +192,7 @@ export class FrameLoop {
         const w0 = Math.sqrt(config.tension! / config.mass!)
 
         const dt = 100 / w0
```
aleclarson: Then rename this to `step`

packages/core/src/FrameLoop.ts
```diff
@@ -62,11 +64,18 @@ export class FrameLoop {
...
       (update && update.bind(this)) ||
-      (() => {
+      ((time?: number) => {
         if (this.idle) {
           return false
         }
 
+        time = time !== void 0 ? time : Date.now()
+        this.lastTime = this.lastTime !== void 0 ? this.lastTime : time
+        let step = time - this.lastTime!
```
aleclarson: Same here, rename `step` to `dt`

aleclarson:
> some of our commits conflicted

FYI: Using `git pull --rebase` instead of `git pull` should let you fix the conflicts without introducing a merge commit. Not a huge deal, since I'll be squashing this before merging, but just wanted to let you know ;)

packages/core/src/FrameLoop.ts
```diff
@@ -189,9 +189,7 @@ export class FrameLoop {
       else {
         velocity = animated.lastVelocity !== void 0 ? animated.lastVelocity : v0
 
-        const w0 = Math.sqrt(config.tension! / config.mass!)
-
-        const dt = 100 / w0
+        const dt = 0.05 / config.w0
```
aleclarson: This changed from `100 / w0` to `0.05 / (w0 / 1000)`. Was not using `0.1` intended?

dbismut:
I'm using Github desktop (roast me)

packages/core/src/FrameLoop.ts
dbismut: Done!

packages/core/src/FrameLoop.ts
dbismut: Done!

packages/core/src/FrameLoop.ts
dbismut: Done!

packages/core/src/FrameLoop.ts
dbismut: Done!

packages/core/src/FrameLoop.ts
dbismut: I think I will stabilize this later, I need to run some more tests. Precision with `0.05` or `0.1` isn't so great compared to analytical.

springperf

As you can see from above, the yellow curves (large steps) move (way) too fast compared to analytical / 1ms Euler  (performance gains seem to be important though on the first run).

dbismut: @aleclarson @drcmda pretty accurate bench I think, results vary depending on the spring config though.

https://jsperf.com/react-spring/16

aleclarson: Nice. Would be great if we had a bar graph that compared different spring configs, but idk how easy that would be to put together.

packages/core/src/FrameLoop.ts
```diff
@@ -159,9 +159,8 @@ export class FrameLoop {
         ? config.initialVelocity[i]
         : config.initialVelocity
 
-      const precision = config.precision || Math.abs(to - from) / 1000 || 0.005
-
-      console.log(precision)
+      const precision =
+        config.precision || Math.min(1, Math.abs(to - from) / 1000)
```
dbismut: ### What it does

Today, precision is used to determine when to rest the spring, ie when velocity is close enough to `0` and current position is close enough to its target position. The higher the precision, the longer the spring comes to rest.

Precision is set by default to `0.005`.

However, animating `opacity` requires usually more precision than animating `translate`. In that case, `0.005` is fine for `opacity` (if you're animating `opacity` from `0` to `1`, the spring would come to rest is when its position is higher than `0.995` and smaller than `1.005`. However, animating a `translate` from `0 to 100` would rest when the position is higher than `99.995` and smaller than `100.005`. This amount of precision is probably exaggerated for a movement of that kind.

This commit makes it so that the precision of the animation would by default depend on the magnitude of the displacement. When animating from `0` to `1` the default precision would be `0.001`. When animating from `0` to `100`, the default precision would be `0.1`.

### Why it matters
The commit tries to rest the spring when no more movement is perceivable by the user. Currently and in some situations where the displacement is significant, `onRest` would fire with a noticeable delay vs the moment where the user _thinks_ the spring is at rest. This commit aims at reducing this perception.

### By the numbers
Animating a spring with `config: { tension: 120, friction: 12, mass: 1 }` from `-250` to `250` would originally take `~1450ms` before `onRest` was fired. With this commit, it only takes `~820ms`.

packages/core/src/FrameLoop.ts
dbismut: @aleclarson ok so here is a basic performance test comparing Euler to Analytical (longer bars are better): https://codesandbox.io/s/goofy-blackwell-66tpc

Results for `step = 0.05/w0`
step1

Results for `step = 0.025/w0`
step2

**🚨&nbsp;This is not _per cycle_ performance: one algorithm might be faster and more accurate, which might result in the spring to come to rest later, and therefore the test to appear slower&nbsp;🚨**

So depending on the step we end up choosing, there will be a significant difference with performance compared to the analytical algorithm. Two things to note:

- The longer the step, the lesser the precision (see previous tool) 
- I'm not sure how interruptible the analytical algorithm is

dbismut: Interestingly, running this on my iPhone gives significantly different results.

image

dbismut: I've also integrated `RK4` algorithm and an option to limit the benchmark to the first 30 cycles. That will actually measure the raw performance of the algorithm per frame.

image

My first conclusion is that the analytical algorithm is really impressive.

packages/core/src/FrameLoop.ts
```diff
@@ -197,8 +197,8 @@ export class FrameLoop {
         const numSteps = Math.ceil(dt / step)
 
         for (let n = 0; n < numSteps; ++n) {
-          const springForce = (-config.tension! / 1000000) * (position - to)
-          const dampingForce = (-config.friction! / 1000) * velocity
+          const springForce = -config.tension! * 0.000001 * (position - to)
+          const dampingForce = -config.friction! * 0.001 * velocity
```
aleclarson: Is this faster?

dbismut: By a infinitesimal fraction (I've seen this elsewhere), but it mostly helps getting rid of the parenthesis :)

aleclarson:
I think this is ready to squash and merge.

@dbismut  @drcmda  WDYT?

dbismut:
> I think this is ready to squash and merge.

I'm not sure just yet: the step chosen here `0.05 / ω` can have significant impact on precision as you can see from the graph below:

image
 
I'm trying to understand how to evaluate the marginal error and compute a step formula that would result in better precision. If we want to leave this aside for later, we should probably stick to `step = 1ms` for now.

Should we also amend tests?

packages/animated/src/AnimatedValue.ts
```diff
@@ -58,8 +56,10 @@ export class AnimatedValue<T = unknown> extends Animated
     if (is.num(this.value)) {
       this.startPosition = this.value
       this.lastPosition = this.value
-      this.lastVelocity = isActive ? this.lastVelocity : undefined
-      this.lastTime = isActive ? this.lastTime : undefined
+      if (!isActive) {
+        this.lastVelocity = null
```
dbismut: @aleclarson any reason why `this.lastVelocity` is reset to `null` and not `undefined`? `null` makes the test below not behave as it should:

```js
 velocity = animated.lastVelocity !== void 0 ? animated.lastVelocity : v0
```

aleclarson: That was a mistake, but my local clone has `animated.lastVelocity == null` to account for that change.

packages/core/src/FrameLoop.ts
```diff
@@ -159,7 +159,7 @@ export class FrameLoop {
         : config.initialVelocity
 
       const precision =
-        config.precision || Math.min(1, Math.abs(to - from) * 0.001)
+        config.precision || Math.min(1, Math.abs(to - from) * 0.001) || 0.005
```
aleclarson: Animations enter the frameloop when `to` and `from` are equal

aleclarson: Ah nevermind actually, this will be possible with #807 because `to` can be a dependency.

dbismut:
FWIW, I've updated the precision simulator to enable clamp and initial velocity. The simple algorithm is a simplified version of the analytical algorithm, that isn't really true to underdamped systems but actually has a better behavior. I've also fixed the clamping on analytical solutions which are now closer to being candidates to replacing Semi Implicit Euler.

mtiller:
A few comments as I'm trying to catch up here.  

First, I didn't quite follow the @dbismut means by analytical.  But let me add an observation and he can comment on whether this is what he was doing.  First, let me introduce some terminology:

With a spring, we can talk about the "unstretched length" of the spring.  This is the natural length of a spring.  Think of a physical spring.  It has a natural length when nobody is pushing or pulling it.  That is the unstretched length.  Normally, a spring has a spring constant (in SI units, that would be Newtons/meter).  The force generated by the spring is normally `F = k(x-x_u)` where `x` is the length of the spring and `x_u` is the unstretched length of the spring.  So when they are equal, the spring transmits no force.

Now here is the thing that @dbismut's comments about analytical got me thinking about.  Let's say that you have a spring where `x` is 10 and `x_u` is 5 and (and, of course, you have some mass and some damping), then you actually know precisely where the spring is going to go.  You don't need to integrate, this is a linear system.  The entire trajectory is known out to infinity from just a single equation.  The equation will basically be of the form:

`x = (x_0-x_u)*e^-(a*(t-t_0))*sin(w_0*(t-t_0))+x_u`

...where `x_0` is the initial displacement of the spring, `a` is the time constant of the decay, `w_0` is the natural frequency, `t` is the current time, `t_0` is the time that the motion started and `x_u` is the unstretched length.  If you look closely, you see that the first term there will got to zero for any a>0.  At the point where the motion starts, all you actually need to compute is `x_0`, `a` and `w_0` (you already know `x_u`, that is the `to` value), make a note of `t0` and then every time you get a request for animation frame, just plug `t` into this equation and it will given you back `x`.

Now, there is one complication.  What happens if the `to` value changes?  Presumably, this gets cascaded down as a result of a props change.  Well, no problem.  Just start over.  This happens much less frequently than requests for animation.  The `x_0` will just be the current value of `x` from the old equation, `a` and `w_0` can be easily recomputed (in my opinion, these should actually **be** the props) and `x_u` is just the latest `to` value.  Just update `t_0` to the time when the props where changed and you have an updated, closed-form solution for the trajectory.

This should be pretty much as fast as is possible.  No more need for every RAF to run the integrator to catch up to current time.  Just compute the closed-form solution for what the current position should be.  No need to compute velocities at all.

Note that you can also easily resolve the issue of when to "clamp" the spring.  To be honest, if you put this in place I'm not even sure if there would be a performance issue with just continuing to use the equation indefinitely.  But if you were worried about it, just find out when the amplitude of the displacement ends up being some arbitrary fraction of the initial displacement.  For example, if you want to stop computing things once you get to 1% of the original displacement, you just need to solve this equation for `t_f`:

`e^-a(t_f-t_0)*=0.01`

...or...

`t_f = -ln(0.01)/a + t_0`

That would then be the magic time at which you could just clamp the position to be `to`.

Now I tried to skim through the comments to understand the latest thinking.  Perhaps this is exactly what @dbismut is doing or proposing.  Sorry if I'm just lecturing on something every already realized.  But if you aren't doing it this way, I think this would be the absolute optimal way to proceed in terms of **both** performance and accuracy.

My apologies that all these comments come so late and if I'm somehow missing something important rendering all this commentary totally irrelevant.  I'll try to stay more plugged in to these comments but trust me when I say lots of other stuff in my inbox is currently feeling neglected.

P.S. - At some point, I picked up the code and actually tried to work on the loop in question.  But every time I messed with it, the tests failed.  So I wasn't quite sure how to put together a PR around this.  How are you guys testing?

dbismut:
Hey @mtiller, great to read you here, all this is because of you in the first place! Before all, I would like to apologize for having mentioned your handle far too much, I'm amazed you were able to navigate through all the threads and make sense of them.

Here are some highlights of where I'm at.

### Minimizing steps for semi implicit Euler

Minimizing steps for SIE is precisely the object of this PR that has been greatly inspired by [your post](https://github.com/react-spring/react-spring/issues/678#issuecomment-498066662). The whole objective was to optimize the algorithm cpu-wise and this objective led me to consider analytical solutions — I was using the _analytical_ term by contrast with _numerical_ (Euler, RK4), I think I read that somewhere.

I tried using a step correlated to the stability coefficient `2/ω` (which I believe is only true to undamped oscillators): this PR currently uses a `step = 0.05/ω` but I quickly ran into perceptible accuracy issues with certain spring configurations. This is why I built a spring algorithm comparison tool and a performance benchmark tool which you'll find below.

### Analytical solution (Apple-like)

What you mentioned in your post above is what I figured out after a few days looking at different spring libraries, which all mention Apple for being the source of this equation, but I realized this is just math after all. Below are two implementations of the equation you mentioned, both claim to be inspired from `CASpringAnimation` in iOS.

Whenever the destination value `to` changes, I'm resetting the animation `elapsedTime`, setting the initial velocity `v_0` to the last computed velocity of the spring, and setting the `x_0` origin to the current position of the spring.

I'm still using difference between the current position and the destination target and velocity to evaluate when resting the spring, is there any benefit in using `t_f = -ln(0.01)/a + t_0`?

What made me pursue the numerical route is that I thought that with a sufficient large iteration step the iteration on each `RAF` frame might be faster than computing exponentials and trigonometric functions.

#### Wobble Analytical

This is stolen from the library [wobble](https://github.com/skevy/wobble/blob/2678d7d0f3605ee447c8c1ac4b69eef7b2cd88cf/src/index.ts#L261), and I guess is the most complete equation, differentiating under damped, critically damped and over damped systems.

#### Simplified Analytical

This is a simplified version of the above equation, used by the library `popmotion`, working well for underdamped  springs (but with a different behavior for overdamped springs, you can check that using the tools below). It also uses a numerical approximation of the velocity instead of the executing the derivative of the position equation (which I suspect is faster).

In my opinion, this is the best candidate in replacing the current algorithm.

### Benchmarking tools

As I mentioned, I built two tools to be able to evaluate both accuracy and cpu performance.

#### Precision benchmark
This one is available at https://spring-perf.surge.sh/

It allows you to visualise the curve of the spring movement on a chart. To initiate a precision chart, just click on the pink square. You can also drag it and it will follow your cursor based on the config set in the GUI in the top right.

It also does some very poor performance analysis that I wouldn't look at too much.

perf
Here are the available options from the GUI:
- **Solution:**  solving algorithm. Euler is semi implicit Euler.
- **Step:** iteration step. When **<= 20** it is used as **milliseconds**, when **>20** it is used as  a fraction of the angular frequency `step * 0.001 / ω`
- **Limit Cycles:** will limit the movement to the first **30 cycles**. This is useful when testing an undamped (frictionless) oscillator which never comes to rest.
- **Tension, friction, mass:** characteristics of the oscillator
- **Velocity:** initial velocity of the object
- **Clamp:** whether to clamp the movement (`-1` is no clamping by default, other values are used as restitution coefficient).

#### Performance benchmark
This benchmark is available at https://66tpc.csb.app/ and the source code is on [codesandbox](https://codesandbox.io/s/spring-algorithm-benchmarks-66tpc).

Here `RAF` is replaced with a `while(!finished)` loop to compare raw performance of the different algorithms. Results may vary depending on the computer and browser you're using. Here are my results on Chrome, MacBook Pro 2017:

image

### How to test

If you want to iterate from there and test different algorithms + having the same precision benchmark tool explained above, I suggest you use the forks I created:

If so, you should do the following:

```
$ git clone https://github.com/dbismut/react-spring.git -b feat/performance
$ cd react-spring
$ yarn
$ git clone https://github.com/dbismut/react-use-gesture-examples.git -b v9-performance examples
$ cd examples
$ yarn
$ yarn start
```

The examples will use the local version of `react-spring`.

aleclarson:
Merged into #808

PR #807, opened 2019-09-07
feat: add Spring class

aleclarson:
Closed in favor of #808


barrymay:
If I'm understanding the change correctly for layman's terms , primary target is to allow async/await-based spring sequential timeline processing of spring changes?

PR #809, opened 2019-09-09
feat: useTransition rewrite

aleclarson:
### ⚠️ More testing required. See demos [here](https://github.com/react-spring/react-spring-examples/pull/17)

This PR is a rewrite of `useTransition` from scratch in hopes of simplifying both the API and the internals. Here's an example of the new API:
```tsx
import {animated, useTransition} from 'react-spring'
const MyComponent = ({ items, ...props }) => {
  // A function is returned, instead of an array.
  const transition = useTransition(items, props)
  // Call the function to render your items.
  // The "values" argument is just like what "useSpring" returns.
  // The "item" argument is the item being transitioned.
  return transition((values, item) => (
    <animated.div style={values}>{item}</animated.div>
  ))
}
```
Once #670 is merged, you can pass a deps array and you'll get the `set` and `cancel` functions returned to you:
```ts
const [transition, update, stop] = useTransition(items, props, [...deps])
```
Also, the `Transition` component has changed. Its `children` prop now takes different arguments:
```tsx
<Transition
  items={items}
  from={{ opacity: 0 }}
  enter={{ opacity: 1 }}
  leave={{ opacity: 0 }}>
  {(values, item) => (
    <animated.div style={values}>{item}</animated.div>
  )}
</Transition>
```

## Notable changes

- The `keys` argument is now the `key` prop
  - Note: The `key` prop is **optional** for primitive items and *mutable* object items
- Returns a `transition` function instead of a `transitions` array
  - Pass a render function to it and you'll get the animated values for each item
  - The element you return has its `key` prop set for you automatically
  - The `phase` isn't exposed to the user anymore
- Now safe to use in concurrent mode
- Unique mode is the new default (and the `unique` prop was removed for simplification)
- Lazy mode is the new default (and the `lazy` prop was removed in favor of `expires`)
- Event props cannot be passed directly (eg: `props.onFrame`), but you can still pass them in specific phases (eg: `props.enter.onFrame`)
- The `reset` prop now uses the `initial` prop if it exists
- The `onDestroyed` prop was removed (was there a good use case for this?)
- The `order` prop was removed

### New `key` prop

Alias: `keys`

When immutable objects are passed as items, they need explicit keys in order to reconnect them with any existing transitions. 

For an array of immutable objects, pass a function that maps over the array and returns the unique id of each item...

```ts
key: item => item.key,
```

... or pass an array of keys. 

```ts
// Using lodash.map
key: _.map(items, 'key'),
```

For a single immutable object, you can pass its key directly (no function required).

```ts
key: item.key,
```

### New `expires` prop

For controlling when deleted items are dismounted after their leave animation finishes. Set to `0` for immediate dismount. By default, dismounting is postponed until (1) next render, or (2) all transitions are resting. Any value above `0` is interpreted as "milliseconds to wait before dismounting is forced".


aleclarson:
## Observations

- All of the `useTransition` demos need to be updated (see [here](https://github.com/react-spring/react-spring-examples/pull/17))

## Questions

- Currently, "expired" transitions are dismounted whenever the component re-renders, even if other controllers are still animating. Is this good or bad? 🤔 
- Currently, there's no mechanism for preventing dismounts altogether. Do any use cases need such a thing?

## Ideas

- **New prop:** Force items to finish entering before they can start leaving?
- **New prop:** Prevent new items from entering until old items have finished leaving?
- **New prop:** Force items to enter/leave one at a time?

guopengliang:
Feature proposal for the new `expires` prop:
1. pass in a single animated value, or an array of animated values, so unmount will wait for these additional animated value(s) to rest, or if they're already at rest.
2. pass in a single Promise, or an array of Promises, so unmount will wait for the Promises to resolve.

A real-world use case for (1) above is when animating an interactive chart. Use a Spring value to animate the chart's x/y axis, while `useTransition` to manage the enter/leave/update of elements on the chart.
```javascript
// Pass the animated value `x` to the `expires` prop of transition,
// so that the unmount event can wait for the X-axis animation to rest,
// instead of in the middle of X-axis animation.
const [{ x }, setAxis] = useSpring(() => { x: 100 });
const transition = useTransition(elements, { ...props, expires: [x] });
```

barrymay:
From the top:
> - The onDestroyed prop was removed (was there a good use case for this?)

There was a good use case for this that I can't see a great way of replacing - the point of the onDestroyed was to know when the leave case was processed and completed.  An example is a modal popup driven by useTransition, where the isShown variable allows it to show/hide via opacity changes et. al..  

In this case, the onDestroyed allowed you a clear way to clean up the div for the modal, after the transition was completely processed.  onRest and onChange really don't give that context

(if there's other ways to know this that would be helpful)

aleclarson:
@barrymay A sandbox would be super helpful 👍

barrymay:
Ask and you shall receive - Here you go ;)

https://codesandbox.io/s/tender-brattain-gl84x

aleclarson:
@barrymay I managed to get that working without the `onDestroyed` prop.

[Sandbox link](https://codesandbox.io/s/frosty-water-dkef4) (will be deleted later)

```patch
From 3014b0bb0b179b47eeb7b44698082de441274e7e Mon Sep 17 00:00:00 2001
From: Alec Larson <alec.stanford.larson@gmail.com>
Date: Tue, 1 Oct 2019 09:33:33 -0400
Subject: [PATCH] fix: avoid using "onDestroyed" prop

---
 Modal.jsx | 115 +++++++++++++++++++++++-------------------------------
 1 file changed, 49 insertions(+), 66 deletions(-)

diff --git a/Modal.jsx b/Modal.jsx
index 140d07d..3567851 100644
--- a/Modal.jsx
+++ b/Modal.jsx
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { css, jsx } from "@emotion/core";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useEffect, useRef } from "react";
 import { createPortal } from "react-dom";
 import { animated, useTransition } from "react-spring";
 
@@ -41,59 +41,17 @@ const modalStyle = css`
 
 const Modal = ({ children, title, onCancel, width, height, shown }) => {
   const parentDiv = useRef(null);
-  const appended = useRef(false);
-  const [show, setShow] = useState(false);
+  useEffect(() => {
+    if (shown && parentDiv.current) {
+      parentDiv.current.focus();
+    }
+  }, [shown]);
 
   const sizeOverride = css({
     width,
     height
   });
 
-  const modalRootRef = useRef(document.getElementById("modal-root"));
-
-  const mainDivRef = useRef(document.createElement("div"));
-
-  const getModalRoot = useCallback(() => {
-    const modalRoot = modalRootRef.current;
-    if (!modalRoot) {
-      throw new Error("No modal-root exists!");
-    }
-    return modalRoot;
-  }, []);
-
-  const transitions = useTransition(show, null, {
-    from: { opacity: 0 },
-    enter: { opacity: 1 },
-    leave: { opacity: 0 },
-    onDestroyed: isDestroyed => {
-      if (isDestroyed && !shown) {
-        const mainDiv = mainDivRef.current;
-        const modalRoot = getModalRoot();
-        if (
-          modalRoot &&
-          mainDiv &&
-          Array.from(modalRoot.childNodes).includes(mainDiv)
-        ) {
-          modalRoot.removeChild(mainDiv);
-        }
-      }
-    }
-  });
-
-  useEffect(() => {
-    if (shown) {
-      const mainDiv = mainDivRef.current;
-      const modalRoot = getModalRoot();
-      modalRoot.appendChild(mainDiv);
-      appended.current = true;
-      setShow(true);
-      parentDiv.current && parentDiv.current.focus();
-      return () => {
-        setShow(false);
-      };
-    }
-  }, [getModalRoot, shown]);
-
   const keyHandler = event => {
     if (shown && event.key === "Escape") {
       event.preventDefault();
@@ -101,31 +59,56 @@ const Modal = ({ children, title, onCancel, width, height, shown }) => {
       onCancel();
     }
   };
-  const modalBody = transitions.map(
+
+  const transitions = useTransition(shown, null, {
+    from: { opacity: 0 },
+    enter: { opacity: 1 },
+    leave: { opacity: 0 }
+  });
+
+  return transitions.map(
     ({ item, key, props }) =>
       item && (
-        <animated.div
-          key={key}
-          style={props}
-          ref={parentDiv}
-          css={modalStyle}
-          onKeyDown={keyHandler}
-          tabIndex={0}
-        >
-          <div className="modalPage" css={sizeOverride}>
-            <div className="modalHeader">
-              <div className="title">{title}</div>
-              <div className="controls">
-                <button onClick={onCancel}>X</button>
+        <ModalBody key={key}>
+          <animated.div
+            style={props}
+            ref={parentDiv}
+            css={modalStyle}
+            onKeyDown={keyHandler}
+            tabIndex={0}
+          >
+            <div className="modalPage" css={sizeOverride}>
+              <div className="modalHeader">
+                <div className="title">{title}</div>
+                <div className="controls">
+                  <button onClick={onCancel}>X</button>
+                </div>
               </div>
+              <div className="modalBody">{children}</div>
             </div>
-            <div className="modalBody">{children}</div>
-          </div>
-        </animated.div>
+          </animated.div>
+        </ModalBody>
       )
   );
+};
+
+const ModalBody = ({ shown, children }) => {
+  const mainDivRef = useRef(document.createElement("div"));
+  const modalRootRef = useRef(document.getElementById("modal-root"));
+
+  useEffect(() => {
+    const modalRoot = modalRootRef.current;
+    if (!modalRoot) {
+      throw new Error("No modal-root exists!");
+    }
+    const mainDiv = mainDivRef.current;
+    modalRoot.appendChild(mainDiv);
+    return () => {
+      modalRoot.removeChild(mainDiv);
+    };
+  }, []);
 
-  return createPortal(modalBody, mainDivRef.current);
+  return createPortal(children, mainDivRef.current);
 };
 
 export default Modal;
-- 
2.19.2
```

barrymay:
Thank you @aleclarson makes perfect sense!  I was trying to do the same with useEffect and I misplaced where the return was happening.  All good from me on removing onDestroyed ;)

barrymay:
@aleclarson sorry to bring you back to this, but in trying to upgrade that modal to your latest useTransition model, but I'm hitting a wall with how you're handling keys here.   

Please check https://codesandbox.io/s/react-spring-canary-modal-2b8c9

In the old version, the keys were kept around during the transition so that the output of the map adopted the key for proper activation/disposal, but in the new canary model, it seems we're responsible for managing our keys. 

Any help would be appreciated.

aleclarson:
@barrymay With this PR, the `key` prop is inserted for you. :)

https://github.com/react-spring/react-spring/pull/809/files#diff-8935b8ea906be6bb0efae9462b4a9e9eR328-R336

aleclarson:
@barrymay Ah, there's a bug it seems. When `shown` is false, it still gets a transition. By default, when the modal finishing fading out, it forces a re-render **only** if all of the other transitions are at rest. Sadly, the transition for `shown == false` isn't finished by that time, which means the modal won't be dismounted until next render (which is never, in this case).

When I have time, I'll add a new prop that lets you say "when an item is *this* value, don't create a transition for it" since you won't be rendering a component for `shown == false` anyway.

For now, you can work around it by forcing a re-render in `onRest`:
```ts
// above useTransition call
const [, update] = React.useState(0);

// in useTransition call
leave: {
  opacity: 0,
  onRest() {
	// the setTimeout call is required because of another small bug
    setTimeout(() => update(i => i + 1), 0);
  },
}
```

aleclarson:
@barrymay Ah, there's a bug it seems. When `shown` is false, it still gets a transition. By default, when the modal finishes fading out, it forces a re-render **only** if all of the other transitions are at rest. Sadly, the transition for `shown == false` isn't finished by that time, which means the modal won't be dismounted until next render (which is never, in this case).

When I have time, I'll add a new prop that lets you say "when an item is *this* value, don't create a transition for it" since you won't be rendering a component for `shown == false` anyway.

For now, you can work around it by forcing a re-render even when transitions are still active:
```ts
useTransition(shown, {
  expires: 0,
  // ...
})
```

barrymay:
@aleclarson great that works with expires for now 🎉  

So for my understanding, the Transition process always created transition (essentially triggerable enter/leave springs) for whatever values it sees (that was no change from the old model).  

But there were always cases (like modals here) where it really didn't need to create a the transition for specific cases (false) here, and due to the added lazy default setting, that false transition never completed.

Is that a fair interpretation?

aleclarson:
> But there were always cases (like modals here) where it really didn't need to create a the transition for specific cases (false here), and due to the added lazy default setting, that false transition never completed.
> 
> Is that a fair interpretation?

Almost. The `false` transition does finish, but then it's failing to check if expired items are still mounted. :)

barrymay:
Got it - thanks!

aleclarson:
Merged into #808

barrymay:
Hi  @aleclarson (let me know if you want a new issue, but this is related to this transition change)

While the expires change here worked great for my simple modal, I'm seeing that the behavior of still seems to have changed in a problematic way for the leave transition.  The new transition (even with expires) seems to keep the spring in "leave" mode and transitions back from that, whereas v8 didn't do that.

Check out these two PR's (using react-three-fiber)
- Line 19 has the useTransition call:
- Click the enter leave button to see the animation.  

It should behave as follows:
- First click: Text comes in
- Second click: Text flies out
- Third click: Text comes in just like it did the first time

(but you'll see in the v9 version the third click using the new model, the text reverses from the third click)

Using v8
https://codesandbox.io/s/react-spring-5-final-demo-1xquj

Using v9 latest canary:
https://codesandbox.io/s/react-spring-5-final-demo-v9-hy6mc

Thanks!

aleclarson:
@barrymay Since `unique` mode was removed and is now the default behavior, transitions are reused when they have the same key. The easiest workaround is to always pass a new object as the first argument:
```tsx
// Always pass a new object as the first argument...
const transition = useTransition({ floatIn }, { ... })
// ...then use destructuring:
transition((style, { floatIn }) => (...))
```
This works because, when the `key` prop is undefined, the value is used as the key, and, in this scenario, the value is always a new object, so no transitions are reused.

barrymay:
Yeah that seems to work to some extent, but then I believe I'll need to manage the key value with the new keys function property, since that {floatIn} object becomes the key itself which doesn't work appears to slow the render down quite a lot (since I'm thinking React is trying to reconcile on an object).  I'm sure I'll be able to work it out, but it's making the interface a bit tougher IMO.

aleclarson:
@barrymay As you can see [here](https://github.com/react-spring/react-spring/blob/41c6f2f604afe8640efd0156d32466f9368f4acc/packages/core/src/useTransition.tsx#L328), the `key` of the element returned by your `transition` render callback is always overwritten by the `transition` function.

barrymay:
you're right.. so the perf issues I was seeing was probably something else.  so based on your code, it looks like you're auto-generating an id for that kind of case? 

If so, It could be helpful to have that fill in the prop key, because that's what I saw when I console logged the key from the transition props.

aleclarson:
> If so, It could be helpful to have that fill in the prop key, because that's what I saw when I console logged the key from the transition props.

Idk what you mean here

barrymay:
It's a bit minor, but more explanatory to an end user if they're looking for the key the lib is using.

If you go here:
https://codesandbox.io/s/react-spring-5-final-demo-v9-kcbbn

I'm logging the key and the ctrl.id in the output, here's the result (object is the key, number is ctrl.id):

```
▶Object {floatIn: false} 4
▶Object {floatIn: false} 4
▶Object {floatIn: true} 5
▶Object {floatIn: false} 4
▶Object {floatIn: true} 5
▶Object {floatIn: false} 6
```

while it's _now_ clear to me that you are getting the actual value from the ctrl.id (which helps), I think showing the {floatIn} object in the key prop is misleading.

Lynges:
Hi @aleclarson I might have discovered a bug.
Playing around with the example for the notification-hub and I am having some trouble restarting the animations without the rest of the queue just rolling on regardless.
This is the example just to be clear: [codesandbox:7mqy09jyq](https://codesandbox.io/embed/7mqy09jyq)

I have created a second button like so
```JSX
    <Button onClick={() => life.reset()} >
    </Button>
```
And if I click it the life animation starts over at the beginning just as expected, but the queued animations still run and thus removes the notification before the restarted life animation is done.

Is this intentional? If so, how might I externally trigger a restart of the life animation such that the notification stays on screen for another round of the timeout duration?

aleclarson:
@Lynges Interesting! Your question is basically: "Is the `reset` method supposed to resolve the promise returned by `await next({ life: '0%' })` or should it stay pending until the reset animation is finished?"

Conceptually, the `next` promise represents a single update to the underlying `Controller` (created by `useTransition` for a specific item), and the `reset` method effectively sends an entirely separate update that happens to interrupt the `{ life: '0%' }` update.

With that in mind, the question becomes "Should interrupting updates extend the lifetime of promises related to the interrupted updates?" and I think the answer is probably "no". The mental model of "when an update-related promise is resolved" gets too muddy if we do that, IMO. And there are definitely work arounds, so it's not a huge issue either. But I'm open to debate here if you have any good points. 👍

Lynges:
I can see why that would cause some problems, but could you point me towards some of those workarounds? :) I have tried setting the `onRest` on the `next` promise that animates `life` to `0%`, but I have not been able to get it working.

Lynges:
The use case is simply that I have an external process that want to change the notification in-place, but when doing so it should get a "new lease of life" in order to give the user a chance to view the new content.

Lynges:
@aleclarson My apologies. Don't know why I thought I was linking a new sandbox. I indeed working from the code you linked.

I cannot prevent it from running the last line in the `enter` callback: `cancelMap.get(item)()`

If I comment out that line, then `reset` appears to reset the animation, but not the duration which results in  the animation running faster to complete at the same point in time. At least that is what appears to be happening.

Lynges:
@aleclarson Just now saw your edit. I tried it, but it still runs `cancel` just as I click the reset button.

`life.pause()` works, as does `life.start()` and `life.stop()`, but if I call any of `update`, `set` or `reset`, the promise returns as finished and `cancel` is called.

aleclarson:
@Lynges Yeah that's a bug I'm looking into now. My suggested edit was still wrong though, for other reasons. Give me a minute to figure this out. :)

aleclarson:
@Lynges Ok, sorry for the wait. Things got a little tricky (and I was distracted by the React Conf stream 😜), but 6 fixes later, everything should work as expected.

Add the following prop to the `next({ life: '0%' })` animation:

```ts
onRest: ({ finished }) => finished && cancelMap.get(item)()
```

Lynges:
Thanks a bunch @aleclarson :) Got it all working. Found a couple of things with regards to when the components are re-rendered. I will do some more digging on Monday to see if it's just my code or something that might be relevant.
Again, thanks a lot for your help and quick response :)

dellwatson:
I still don't understand how to toggle with new useTransition 

`const [toggle, set] = useState(false)
const transitions = useTransition(toggle, null, {
from: { position: 'absolute', opacity: 0 },
enter: { opacity: 1 },
leave: { opacity: 0 },
})

return transitions.map(({ item, key, props }) => 
item
  ? <animated.div style={props}>😄</animated.div>
  : <animated.div style={props}>🤪</animated.div>
)`

dimroth:
@dellwatson Hi, the following should work if you pass the proper index but there might be better options : 
```javascript
const items = [😄, 🤪]
```


```javascript
const [index, set] = useState(0)
const transitions = useTransition(items[index], {
from: { position: 'absolute', opacity: 0 },
enter: { opacity: 1 },
leave: { opacity: 0 },
})

return transition((values, item) => <animated.div style={values}>{item}</animated.div>
)
```

lorenzocadamuro:
I noticed that items are not removed when the animation ends.

Am I missing something?

https://codesandbox.io/s/late-wave-nm2me

Furthermore since the `useTransition` result  is a function, I can't access to the transitions array anymore – I know it sounds strange, but there're situations where I need to share the transitions with other components outside.

Inviz:
Does `immediate: true` do anything in this version? Can't get it to work, it feels like it's just ignored.

aleclarson:
@Inviz Which version are you using?

rayandrews:
Hi @aleclarson, I am trying the latest canary.

My question : is there any effect if I set the `reverse` option in `useTransition`?

I want to reverse the transition animation from last item to first item instead of first item to last item like usual and it seems that option is not working at all

Thank you in advance!

PR #823, opened 2019-10-03
Use Array.from project param instead of map

topaxi:
This is a microoptimization in file size and runtime. `Array.from()` supports a project argument as the second parameter to map each value, therefore it is not necessary to use `.map()` right afterwards.

Feel free to close if you think this might hurt expliciteness or readability.


aleclarson:
Thanks for the PR! This isn't applicable to the `v9` branch, so I'm closing.

PR #825, opened 2019-10-04
 Use Array.from project param instead of map

topaxi:
Originally https://github.com/react-spring/react-spring/pull/823 but for `v9`.

This is a microoptimization in file size and runtime. `Array.from()` supports a project argument as the second parameter to map each value, therefore it is not necessary to use `.map()` right afterwards.

Feel free to close if you think this might hurt expliciteness or readability.


aleclarson:
I hate to do this to you again... but the changes you made aren't applicable to #809, which will be merged into v9 for the next beta. 😅

PR #841, opened 2019-10-26
Updated prop-types and runtime form babel

pawelgraczyk:


aleclarson:
What does this fix? They already use caret ranges

PR #843, opened 2019-10-28
fixes types for setSprings

robbiespeed:
Rather than accept a number and return an object the update function should accept a callback and return void


aleclarson:
v8 is not maintained. Please try v9: `npm install react-spring@next`

PR #854, opened 2019-11-09
fix type: use imported RefObject

xiel:
I was running into an typescript error when doing a production build in a preact project:
```
ERROR in [at-loader] ./node_modules/react-spring/web.d.ts:107:15
    **TS2694: Namespace 'React' has no exported member 'RefObject'.**
```
Fix: Use the imported RefObject (like in all other places in the file).

Amazing project and great demos btw!


joshuaellis:
Does this change need to be moved to v9? We're focussing efforts on that right now.

PR #860, opened 2019-11-16
Update formula for frequency response

aholachek:
Hi, this is using the formula I found [in this article](https://medium.com/ios-os-x-development/demystifying-uikit-spring-animations-2bb868446773), which seems consistent with  [apple's guidelines](https://developer.apple.com/videos/play/wwdc2018/803/?time=2117). I used it to make [this interactive demo](https://spring-playground.netlify.com/). 

I wasn't sure whether to add more logic around guarding against invalid values, please let me know if that would be preferred.

(This pr was made in response to [this comment by @aleclarson](https://github.com/react-spring/react-use-gesture/issues/109#issuecomment-554409570).)

### DIFF
--- a/packages/core/src/Animation.test.ts
+++ b/packages/core/src/Animation.test.ts
@@ -1,0 +1,19 @@
+import { AnimationConfig } from './Animation'
+
+const config = new AnimationConfig()
+
+describe('AnimationConfig', () => {
+  describe('merge', () => {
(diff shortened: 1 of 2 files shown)


aleclarson:
This is perfect, thanks for contributing. I don't remember where I got the original formula. 😅 

> I wasn't sure whether to add more logic around guarding against invalid values, please let me know if that would be preferred.

Yeah sure why not!

PR #866, opened 2019-11-19
Backport skipAnimation global

### LINKED ISSUE #668: Is there a way to global disable transitions?
## 🤓 Question
We are using cypress to ui testing are facing problems with headless browser & animations. Is there a way to global disable so we can add a parameter to our build?

Thank you.

topaxi:
Based on 98b1649ea2d48323ea2b903d94d8584a594fe990, fixes #668. 

While I realize that v9 is on the way, we have some trouble getting screenshot/visual tests going while migrating some other animation libraries to the stable version of react-spring. I'm reimplementing the v9 disable mechanism to the old API for a way to use this feature immediately until the new version is released.

Would be great if we could get this backport into the version 8 branch/release.

### DIFF
--- a/src/renderprops/animated/Controller.js
+++ b/src/renderprops/animated/Controller.js
@@ -1,7 +1,7 @@
 import Animated from './Animated'
 import AnimatedValue from './AnimatedValue'
 import AnimatedArray from './AnimatedArray'
-import { now, colorNames } from './Globals'
+import { now, colorNames, forceImmediate } from './Globals'
 import { addController, removeController } from './FrameLoop'
 import {
   interpolateTo,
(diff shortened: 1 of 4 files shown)


raunofreiberg:
Would love to see this merged. Having trouble testing components with animations due to them not unmounting in tests.

src/renderprops/animated/Controller.js
```diff
@@ -51,6 +51,11 @@ export default class Controller {
     if (reverse) {
       ;[from, to] = [to, from]
     }
+
+    if (skipAnimation) {
+      immediate = true
```
aleclarson: This isn't equivalent to the current implementation (which exists in #808):

https://github.com/react-spring/react-spring/blob/5e54f7dad6701f4179e2f16974a16a4e39840f29/packages/core/src/SpringValue.ts#L776-L778

Because it fails to avoid calling `requestAnimationFrame`.

aleclarson:
> Having trouble testing components with animations due to them not unmounting in tests.

@raunofreiberg I don't understand what you mean

topaxi:
Mocking rAF isn't an option as we are running this in an end-to-end test. I'll see whether I can update it to not call rAF aswell, not an issue in our case but I can see that it would make sense for component level tests.

Issue #903, opened 2020-01-17
useTransition: How to start / stop; and performance issue


chenesan:
I'm building bar chart race animation with useTransition. here is the [source](https://github.com/chenesan/bar-chart-race) and [demo](bar-chart-race.yishan.toys). I have the following questions:

1. I'd like to build a play / stop button for it. But I couldn't find a way to play / stop the transition in the document. Is there any way to do it?

2. The animation is janky on firefox android(at least on my device) but smooth on chrome android. I found out (with react profiler) that render [`<RacingBarGroup>`](https://github.com/chenesan/bar-chart-race/blob/master/src/RacingBarGroup.js) (which call `useTransition` for animating the bars) cost lots of time. Is there any idea for optimizing performance?

    I noticed that the other bar chart race implementation using vanilla js and d3 is smooth on firefox android (for example https://observablehq.com/@d3/bar-chart-race), so I'm not sure this issue is due to my code, react-spring or react.

I previously posted this on [spectrum](https://spectrum.chat/react-spring/general/building-a-bar-chart-race-and-get-some-questions-with-usetransition~8963256b-a362-49e4-9278-957ceb451f53), but I got no answer there so I created a new issue here. Sorry for bombing if you've read my questions on spectrum :-/

aleclarson:
There *is* a pausing mechanism in [the latest canary](https://spectrum.chat/react-spring/general/try-the-new-v9-canary~c5cad1eb-c6ff-4b40-9a10-76d368c3e404) (see the `SpringValue#pause` and `SpringValue#resume` methods), but it's not integrated with the `useTransition` API yet. Eventually, it will be, and if you have time to implement it, you can open a PR that adds `pause`/`resume` methods to the `useTransition` ref API [here](https://github.com/react-spring/react-spring/blob/162b2be9c75c153045826d757c22533bb36ad616/packages/core/src/useTransition.tsx#L293-L317). Of course, you'll need to get acquainted with the internals to do that, and I try to keep the source code relatively understandable. Hope that helps. ✌️

chenesan:
Thanks for the information! That helps a lot. I will try to look into this and maybe make a PR in recent days. Which branch should I work on?
And, how about the perfomance issue? Is there any performance tip for `useTransition`?

aleclarson:
The [`feat/spring-class`](https://github.com/react-spring/react-spring/tree/feat/spring-class) branch holds the latest changes and should be stable. You can build your PR on that.

[...]

PR #910, opened 2020-01-24
Bugfix: make `animated` work with arrow function component

chenesan:
I found out this bug when I'm working on #903: When we pass arrow function component, it will throw error because in `createAnimatedComponent` we check `Component.prototype.isReactComponent` to check if it's component created by `create-react-class`, but arrow function doesn't have `.prototype` property. This commit additionally check if `.prototype` exists to make arrow function component can work with `animated()`.

To test this just try to pass an arrow function component into `animated()`.

### DIFF
--- a/packages/animated/src/withAnimated.tsx
+++ b/packages/animated/src/withAnimated.tsx
@@ -37,7 +37,8 @@ const createAnimatedComponent = (Component: any) =>
     const hasInstance: boolean =
       // Function components must use "forwardRef" to avoid being
       // re-rendered on every animation frame.
-      !is.fun(Component) || Component.prototype.isReactComponent
+      !is.fun(Component) ||
+      (Component.prototype && Component.prototype.isReactComponent)
 
     const forceUpdate = useForceUpdate()
     const props = new AnimatedProps(() => {


aleclarson:
> arrow function doesn't have `.prototype` property

Wow, that's very weird.. 🤯

packages/animated/src/withAnimated.tsx
```diff
@@ -37,7 +37,8 @@ const createAnimatedComponent = (Component: any) =>
     const hasInstance: boolean =
       // Function components must use "forwardRef" to avoid being
       // re-rendered on every animation frame.
-      !is.fun(Component) || Component.prototype.isReactComponent
+      !is.fun(Component) ||
+      (Component.prototype && Component.prototype.isReactComponent)
```
aleclarson: Let's use `Object.getPrototypeOf(Component).isReactComponent` instead

chenesan: Nice fix. Fixed and force push c4d4cf0.
Buy the way I learned that arrow function do have prototype `.__proto__`, which is `Function`. But it doesn't has `.prototype` since it's not a constructor. Javascript is really weird..

aleclarson: Oh, of course. Not sure how I missed that 😆

aleclarson: I confused myself, so my suggestion didn't make sense. Please revert the newest commit. 😅

chenesan: Uhhh I didn't think of the original `create-react-class` case 😓 OK revert it

chenesan: Fixed and force push a2d71a7

aleclarson:
Thanks @chenesan 🎉

PR #911, opened 2020-01-26
[Feature] Add resume and pause method into useTransition/useSpring ref api

### LINKED ISSUE #903: useTransition: How to start / stop; and performance issue

chenesan:
Add `pause` and `resume` method in `Controller`, `useSprings` and `useTransition` ref api. I have tried it on my own project and it looks great. If needed I can make a minimal demo.

This change will close #903 .

### DIFF
--- a/packages/core/src/Controller.ts
+++ b/packages/core/src/Controller.ts
@@ -242,6 +242,28 @@ export class Controller<State extends Indexable = UnknownProps>
     return this
   }
 
+  /** Freeze the active animation in time */
+  pause(keys?: OneOrMore<string>) {
+    if (is.und(keys)) {
+      this.each(spring => spring.pause())
+    } else {
+      const springs = this.springs as Indexable<SpringValue>
+      each(toArray(keys), key => springs[key].pause())
+    }
+    return this
+  }
+
+  /** Resume the animation if paused. */
+  resume(keys?: OneOrMore<string>) {
+    if (is.und(keys)) {
+      this.each(spring => spring.resume())
+    } else {
+      const springs = this.springs as Indexable<SpringValue>
+      each(toArray(keys), key => springs[key].resume())
+    }
+    return this
+  }
+
   /** Restart every animation. */
   reset() {
     this.each(spring => spring.reset())
--- a/packages/core/src/types/spring.ts
+++ b/packages/core/src/types/spring.ts
@@ -177,6 +177,30 @@ export type SpringStopFn<T> = T extends object
   : () => void
 
 /**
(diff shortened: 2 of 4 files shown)


packages/core/src/Controller.ts
```diff
@@ -242,6 +242,28 @@ export class Controller<State extends Indexable = UnknownProps>
...
+  /** Freeze the active animation in time */
+  pause(keys?: OneOrMore<string>) {
+    if (is.und(keys)) {
+      this.each(spring => spring.pause())
+    } else {
+      const springs = this.springs as Indexable<SpringValue>
+      each(toArray(keys), key => springs[key].pause())
+    }
+    return this
+  }
```
chenesan: I intentionally copy and paste same logic from `stop` method since the logic may be different in the future.

packages/core/src/Controller.ts
```diff
@@ -242,6 +242,28 @@ export class Controller<State extends Indexable = UnknownProps>
...
+  resume(keys?: OneOrMore<string>) {
+    if (is.und(keys)) {
+      this.each(spring => spring.resume())
+    } else {
+      const springs = this.springs as Indexable<SpringValue>
+      each(toArray(keys), key => springs[key].resume())
+    }
+    return this
+  }
+
```
chenesan: Same as above

aleclarson:
Looks good 👍 

Note that the pausing feature doesn't yet support the `delay` prop.

PR #915, opened 2020-01-27
feat: support the ref prop to useTrail

1natsu172:
Issue #838

Support the ref prop to `useTrail` like the `useTransition`.

> useTrail with useChain at my local

PR #918, opened 2020-01-29
feat: support ref prop to useTrail on v9 canary

1natsu172:
Port #915 to v9 canary 🚢


aleclarson:
Closing in favor of #1134

PR #940, opened 2020-02-26
Fix: HSL -> RGB Accuracy. 

ShanonJackson:
https://github.com/react-spring/react-spring/issues/938

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -19,16 +19,16 @@
     "gzipped": 1999
   },
   "dist/web.js": {
-    "bundled": 63554,
-    "minified": 30170,
-    "gzipped": 10988,
+    "bundled": 63781,
+    "minified": 30180,
+    "gzipped": 10974,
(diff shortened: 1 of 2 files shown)


aleclarson:
Merged into #978

aleclarson:
Now available in v9.0.0-rc.2 #985

aleclarson:
This PR will be reverted in v9.0.0-rc.4 because it returns `rgb(255, 97, 97)` for `hsl(57, 100%, 69%)` when it should actually be `rgb(255, 247, 97)`. The previous implementation works as expected, at least in this case.

ShanonJackson:
Might be worth just using the native DOM api to convert colors, it (should) always be right. However this wont work for SSR.
```
var div = document.createElement("div");
div.style.color = "hsl(57, 100%, 69%)"
console.log(div.style.color) // rgb(255, 247, 97);
```
where it always converts to RGB.

PR #945, opened 2020-03-13
Fix memory leak in SSR

topaxi:
This is an ugly hack to fix a major memory leak in our application. Without this our application crashes with OOM errors after about 5mins.

I'd be very happy to change the solution to something more appropriate if you can point me into the right directions, but so far I'm a bit lost where the controllers should be cleaned up in the SSR case.

Related to https://github.com/react-spring/react-spring/issues/778

### DIFF
--- a/src/animated/FrameLoop.ts
+++ b/src/animated/FrameLoop.ts
@@ -5,6 +5,10 @@ import { now, requestFrame, manualFrameloop } from './Globals'
 let active = false
 const controllers = new Set()
 
+if (typeof window === 'undefined') {
+  setInterval(controllers.clear.bind(controllers), 1000)
+}
+
 const update = () => {
   if (!active) return false
   let time = now()


KB1RMA:
As I'm reading through this file, wouldn't it maybe be more appropriate to not add a controller to the set at all when server rendering? The animation is never going to run on the server, so why add it at all?

I don't know if this is the best place to add it, but checking if we're on the server when start is called could prevent the leak. Something along the lines of this:

```js
const start = (controller: Controller) => {
  // Check for SSR
  if (typeof window === 'undefined') return

  // Proceed as normal in browser
  if (!controllers.has(controller)) controllers.add(controller)
  if (!active) {
    active = true
    if (manualFrameloop) requestFrame(manualFrameloop)
    else requestFrame(update)
  }
}
```

Does this approach seem reasonable?

joshuaellis:
this looks the same as #1054 is there a difference?

Anthonyzou:
SSR memory leaks seem fixed in version `9.0.0`.

PR #980, opened 2020-05-03
fix: useTrail v9

aleclarson:


packages/core/src/hooks/useTrail.ts
```diff
@@ -70,17 +68,17 @@ export function useTrail(
     }
   }, deps)
 
-  if (arguments.length == 3) {
+  if (propsFn || arguments.length == 3) {
     const update = result[1]
-    result[1] = propsArg => {
+    result[1] = useCallbackOne(propsArg => {
```
aleclarson, line 73: This call means you must **never** switch between passing a props function and a props object.

PR #983, opened 2020-05-04
Add esm support to shared package

TrySound:
Enabled tslib to prevent emitting helpers with top level `this` which
does not exist in es modules. For some reason tslib has different and
correct helpers.

Commonjs artifacts are moved into cjs folder. Added esm folder near it.
package.json routes are created for each module via cherry-pick package.

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -4,16 +4,18 @@
   "description": "Globals and shared modules",
   "main": "src/index.ts",
   "scripts": {
-    "build": "tsc -p ."
+    "build": "tsc -p . --outDir dist/cjs && tsc -p . --module ES6 --outDir dist/esm && cherry-pick --cwd=dist ../src --cjs-dir=cjs --esm-dir=esm"
   },
   "publishConfig": {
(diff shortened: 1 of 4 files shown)


TrySound:
Have no idea why tests failed.

aleclarson:
Thanks!

`cherry-pick` is a neat little tool 😮

aleclarson:
I noticed a `./dist/createInterpolator.test/` directory is created. Not a big deal, but I wonder if there's a way to avoid that.

TrySound:
Maybe we can add the feature to cherry-pick. Do you mind to release current changes? We have to alias local copy of "shared" package to build spring with rollup.

aleclarson:
Nested directories aren't supported by `cherry-pick` either. https://github.com/Andarist/cherry-pick/issues/8

So the `src/types` directory isn't being copied.

TrySound:
Types could be reexported with import type {} syntax to prevent bundle bloating.

aleclarson:
> Do you mind to release current changes?

#985

PR #1022, opened 2020-05-23
Parallax math fix: Remove targetScroll

localjo:
This is a proposed solution for https://github.com/react-spring/react-spring/issues/962

I apologize if this PR is not fully tested or complete or if the base branch is wrong. I [couldn't find any documentation](https://github.com/react-spring/react-spring/issues/1021) about testing and contributing.

But I can say this fix works for me.

PR #1023, opened 2020-05-23
fix(parallax): remove extra targetScroll space

### LINKED ISSUE #692: [TypeScript] Missing scrollTop and scrollLeft typings
## 🐛 Bug Report

`scrollTop` and `scrollLeft` are missing in `animated.*`

https://codesandbox.io/s/inspiring-cookies-grx45

## Environment

- `react-spring` v8.0.20, 9.0.0-beta.8
- `react` v16.8.4
- `typescript` v3.5.1, v3.3.3

### LINKED ISSUE #962: Dead space when using ParallaxLayer speed property
## 🐛 Bug Report

Using an ParallaxLayer with a non-zero `speed` creates a dead space in which it is impossible to position the layer (no matter its `offset` value).  
It seems to me that it is a maths problem occuring [here](https://github.com/react-spring/react-spring/blob/66bda492990f2bcd38225b5fe22997a92144c249/src/renderprops/addons/Parallax.js#L62) and [here](https://github.com/react-spring/react-spring/blob/66bda492990f2bcd38225b5fe22997a92144c249/src/renderprops/addons/Parallax.js#L48) (more explanations in the repro).

## To Reproduce

See the explained codesandbox demo below.

## Expected behavior

Being able to vertically align a layer at any height, not matter its `speed`.

[https://codesandbox.io/s/withered-sun-zqbh3](https://codesandbox.io/s/withered-sun-zqbh3)

Seeing the behaviour requires you to open the sandbox as a full page app [(https://zqbh3.csb.app/)](https://zqbh3.csb.app/)

> I will explain the repro using a screen height of 1000px, but the explanation is valid for any screen height, only the dead space size will vary (the screen width does not matter at all, at least for vertical parallax).

The repro has one 2-pages Parallax element, with 3 ParallaxLayers inside - yellow, cyan and blue - each having a `speed` of **0.1**.
> as for the screen height, changing the speed changes the dead space size.

With this configuration, the layers original positions are at:
- yellow: `transform: translate3d(0px, 900px, 0px);`
- cyan: `transform: translate3d(0px, 1000px, 0px);` (the browser rounds this value to an entire pixel, the actual result should be 999.999999)
- blue: `transform: translate3d(0px, 1100px, 0px);`

**The consequence of this is:** I am able to create layers between yellow and cyan (i.e. between 900px and 1000px) with offsets from 0.9 to 0.9999...99999 **BUT** I am unable to create layers between cyan and blue (i.e. between 1000px and 1100px) because there is no actual value between 0.9999...99999 and 1.

The code responsible for translating `offset` and `speed` of a layer to an actual height in pixels is the following (see links above):
```javascript
const targetScroll = Math.floor(props.offset) * parent.space
const offset = parent.space * props.offset + targetScroll * props.speed
const to = parseFloat(-(parent.current * props.speed) + offset)
```
Seems to me that `Math.floor` is the problem. Removing it _seems_ to remove the dead space (⚠️ _seems_ from what I tested, and I am definitely not very confident with maths).  
The codesandbox has a `parallax.js` which does not contain any `Math.floor` for you to try.

## Environment

- `react-spring` v8.0.27
- `react` v16.13.1

> **Note:** sorry for the long text but the issue is quite touchy and not easy to actually visualize. I hope I made it clear 🙃

localjo:
this was causing some elements to have an extra viewport height gap. see issue for more details.

fixes #962


localjo:
This replaces #1022 which was applied to the wrong branch.

localjo:
It appears that the build is failing due to `yarn --frozen-lockfile` failing. I don't think that is related to this change, so I'm guessing it's an issue with Travis.

joshuaellis:
Can you merge current version 9 into your branch please so we can review this properly, I think somethings gone wild with the commits. Thank you ⭐

joshuaellis (approved):
this has been a long time coming, but thank you for your contribution ⭐

PR #1029, opened 2020-05-26
feat(useTransition): when `keys` is null, generate keys internally

aleclarson:
Fixes #1027

Use `key: null` or `keys: null` to force duplicate items to have a separate `TransitionState` object. When used, `TransitionState` objects in the "leave" phase are **never** reused.

- [ ] Write tests


aleclarson:
Merged into v9 manually: https://github.com/pmndrs/react-spring/pull/632/commits/bdecdc15584da200a40571ae7eb6c96a8186f78a

*Still needs unit test though...*

PR #1126, opened 2020-08-26
Correct SetUpdateCallbackFn types

rockwotj:
Currently the example from https://www.react-spring.io/docs/hooks/use-springs does not work in typescript.


jakeblaxon:
Can we get this merged in? This issue is currently affecting us and it'd be nice to get this fix in.

Issue #868, opened 2019-11-20
useTrail Does Not Pass Index.


Tenkir:
## 🐛 Bug Report

`useSprings` passes the current index when passed a function. `useTrails` does not. This prevents the modification of elements by array index.
 
## To Reproduce

***Example:***
```
const props = useTrail(items.length, (i) => ({
  from: { opacity: 0 },
  opacity: i * .25
}));
```

***Result:*** `i` is undefined, therefore target `opacity` is always `0`.

## Expected behavior

`i` is the current index being mapped.

## Link to repro
https://codesandbox.io/s/react-spring-usetrails-index-bug-ci0t6

## Environment

- `react-spring` v8.0.27
- `react` v16.11.0

PR #1131, opened 2020-09-04
Adds injection of index param on useTrail's functional API

jdmg94:
This commit preserves current behaviour of useTrail and adds the injection of the index parameter on the functional API of the hook matching the funcitonal API of useSprings, sans Controller injection.

fixes issue #868


aleclarson:
Already fixed in v9

https://github.com/pmndrs/react-spring/blob/c0bbbe0da1866c7d1bfcd0b42711ad2acd7815ab/packages/core/src/hooks/useTrail.ts#L51

Issue #1008, opened 2020-05-14
[9.0.0-rc.3] Event props in useSpring not containing full list of values


tim-soft:
## 🐛 Bug Report

Following up on https://github.com/react-spring/react-spring/issues/997 

In previous releases, `f` from `onChange` would be an object containing all animated values at that instance

```jsx
const [props, set] = useSprings(pages.length, i => ({
    x: i * window.innerWidth,
    scale: 1,
    onChange: f => console.log(f),
}))
```

i.e. `f` in this case would normally be 
```
{
    x: 1234,
    scale: 1.2
}
```

In v9.0.0-rc.3, f is equal to whatever value has changed in that frame
```
// Scale changes 1.0 -> 1.2
f = 1.2
// x changes 1234 -> 1235
f = 1235
```

 I'm not sure if this change was intentional or not but I think the old behavior is much more desirable/useful

aleclarson:
Fixed by https://github.com/react-spring/react-spring/commit/25da5024e659af9462705018c4fc80c61c2302fc, will be included in `9.0.0-rc.4` when released.

More tests are definitely needed if something like this is slipping through. 🙃

Issue #937, opened 2020-02-24
Animations not working on inline svg v9


loren-jiang:
## 🐛 Bug Report

Animations not working on inline svg components in v9. Works in v8.

## Link to repro (highly encouraged)

[ demo](https://codesandbox.io/s/react-spring-inline-svg-grokm)

## Environment

- `react-spring` v9.0.0-beta.34

aleclarson:
Please try with [the latest canary](https://spectrum.chat/react-spring/general/try-the-new-v9-canary~c5cad1eb-c6ff-4b40-9a10-76d368c3e404). 👍

Issue #929, opened 2020-02-05
Parallax is not working anymore


MBach:
## 🐛 Bug Report

Parallax is not working anymore

## To Reproduce

Update     
`"@react-spring/parallax": "9.0.0-canary.808.11.4be2ac3"`
to
`"@react-spring/parallax": "9.0.0-canary.808.17.55c5691"`

With
```
<Parallax ref={parallax} pages={2} enabled={false}>
...
</Parallax>
```

## Expected behavior

Layer should change when one is dragging the DIV.

## Link to repro (highly encouraged)

https://codesandbox.io/s/react-spring-parallax-qt4tr

## Environment

- `react-spring` c9.0.0-canary.808.17.55c5691
- `@react-spring/parallax` v9.0.0-canary.808.17.55c5691
- `react` v16.12.0

aleclarson:
Please try with v9.0.0-rc.2 #985

If the bug persists, please give a step-by-step list on what to do in your repro. 👍

Issue #922, opened 2020-02-01
Stale components are not removed properly


Slapbox:
## 🐛 Bug Report

When re-rendering a list and adding a children, the items are added after the existing elements in the DOM, the ones that should be removed. They're transitioned out, and something is getting removed, but not everything that needs to be 

## To Reproduce

[Repro](https://codesandbox.io/s/nice-bouman-16quc)

## Expected behavior

![image](https://user-images.githubusercontent.com/6835891/73598848-2b33e700-450b-11ea-92b2-46eed7ad3d78.png)

Elements render in-place, rather than each item addition resulting in a new set of entries _below_ the existing.

## Link to repro (highly encouraged)

[Repro](https://codesandbox.io/s/nice-bouman-16quc)

## Environment

- `react-spring` `v9.0.0-808.17`
- `react` `v16.12.0`

amadeus:
I'm not sure you are using transition group properly. You never remove anything from the array of items - therefore nothing ever gets removed?

amadeus:
There's also an issue where you don't key items in the array you pass in. If you do this: https://codesandbox.io/s/epic-villani-p1dn8 then at least you start getting chunks removed

Slapbox:
@amadeus the code you provided does work better, but left running alone long enough it exhibits roughly the same behavior. 

If you force `opacity: 1` for all the `li` elements you'll see:

![image](https://user-images.githubusercontent.com/6835891/79025139-ece22600-7b52-11ea-8ecc-600642cc20e0.png)

Pretty sure it's related to this same repro (provided by @aleclarson in another thread): https://codesandbox.io/s/react-spring-issue-927-bjrbv

I could certainly be doing something wrong. Writing code using a pre-release version means no documentation means no certainty. I don't _think_ keys need to be passed the way you are, but I'm going off my memory from ~65 days ago. Am I wrong about the need to pass keys?

In `v8` we needed something like `keys={item => item.key}` to define the keys, but as far as I'm aware it's not needed for `v9` (which admittedly does sound wrong.)

aleclarson:
Changing your `componentDidMount` method fixed the issue:
```diff
   componentDidMount() {
     setInterval(() => {
-      const { items, keyToAdd } = this.state;
-      this.setState({ keyToAdd: keyToAdd + 1 });
-      this.setState({ items: items.concat({ key: keyToAdd }) });
+      this.setState(({ items, keyToAdd }) => {
+        return {
[...]

PR #1330, opened 2021-03-18
[v9] Release

joshuaellis:
It's been a while since a stable release. There have been some setbacks, but I think our goal should be to launch v9 as stable and fix from there. We've had 3 `rc` releases and they've been great. 

Hopefully with the release of a stable version people will have the confidence & motivation to fix bugs that they're experiencing for themselves to make this library stronger 💪🏼 

There's 2 small things we need to do first, but this way I can track which issues should be resolved with this release and attach them accordingly.

🎉


joshuaellis:
* resolves #922 
* resolves #929
* resolves #937
* resolves #1008

aleclarson:
DM me on Twitter please. Thanks!

Issue #1185, opened 2020-10-31
scrollTop is not set when `from` & `to` are equal


0x6431346e:
from @aleclarson on discord:

> You need to provide a start value: `from: { scrollTop: 0 }`

This fixes it, yes, thanks! 😄 (I guess on v8.0.27 the lib used to do that automatically.)

To achieve my behavior of a initial scrollTop I have to do this now:

```
  const { scrollTop } = useSpring({
   from: { scrollTop: 0 }
   to: { scrollTop: 100 },
   immediate: true
});
```
And then to set `immediate: false` on subsequent calls.

I can live with this... **hovever** I would still consider this a bug, and expect this expression to work as an equivalent, just like it does in `v8.0.27`:

```
  const { scrollTop } = useSpring({
    from: { scrollTop: 100 }
  });
```

and this should be possible too:

```
  const { scrollTop } = useSpring({
    from: { scrollTop: 100 }
    to: { scrollTop: 100 }
  });
```

It works with `animated.span` so this is most likely an issue in `animated.div`'s scrollTop handling:
https://codesandbox.io/s/react-spring-v9-scrolltop-bug-forked-7h3vf?file=/src/App.js

_Originally posted by @0x6431346e in https://github.com/pmndrs/react-spring/issues/985#issuecomment-719952705_

Rohanhacker:
I am working on this one @joshuaellis

joshuaellis:
resolved by #1354

PR #1354, opened 2021-03-21
fix: scrollTop is not set when from & to are equal #1185

Rohanhacker:
#1185

I think the issue was that we were not calling applyAnimatedValues on the first render (which is getting called on the first render in v8) and thus scroll top was not working. I was able to reproduce and fix it on my local.


joshuaellis (approved):
Thanks for the contribution ⭐

PR #1357, opened 2021-03-23
fix: fix message channel error

Aslemammad:
As we talked in discord, for some machines, testing doesn't have MessageChannel API, so I pollyfilled it and removed the error message to work properly.


joshuaellis (approved):
Thanks for the contribution ⭐

PR #1386, opened 2021-03-31
Chore: remove external deps and add examples

dbismut:
### What it does
- Removes `fluids` and `alloc/types` deps for better maintenance.
- Better structure of the demo sandboxes.


joshuaellis:
We could probably remove `react-layout-effect` although it might be annoying to replace. Basically it swapes out `useLayoutEffect` for `useEffect` if you're SSR.

dbismut:
Right! I've added it to `react-spring/shared` inside the `hooks` directory.

dbismut:
Wow, I had no idea that Codesanbox would trigger ci for sandbox links posted in the first comment 😅

joshuaellis (approved):

PR #1395, opened 2021-04-04
[fix] Event API unification based on useTransition

### LINKED ISSUE #1041: [9.0.0-rc.3] useTransition cancel doesn't work
## 🐛 Bug Report

I forked [MessageHub](https://codesandbox.io/embed/7mqy09jyq), an older react-springs version example demonstrating useTransition. I updated the react-springs version to 9.0.0-rc.3 and attempted to rework the example using the v9 docs. 

Here is my repo: https://codesandbox.io/s/messagehub-150hm?file=/src/index.js

## To Reproduce

Click to create a notification and then click the cross icon on the notification to remove the notification before the leaving animation completes.

## Expected behavior
The notification animation should be cancelled and removed, as it does in the original messagehub example.

## Link to repro
https://codesandbox.io/s/messagehub-150hm?file=/src/index.js
The cancel function is called on line 39.

## Environment

- `react-spring` v9.0.0-rc.3
- `react` v16.8.x as well as `react` 16.13.1

### LINKED ISSUE #1372: Parameters missing from useTransition’s onRest
## 🐛 Bug Report

For examples like the notification center you can find on react-spring.io the onRest callback is used to cleanup after an animation, on v8 an issue was reported where the behavior of this callback was different on useTrail (see #1131 ) this PR was rejected with the reasoning that it was fixed in v9 (months before the release of v9)  up until RC3, this was true, however the API has changed and there is no documentation that reflects the new API changes, on RC the callback for onRest for the useTransition hook had 2 passed arguments, but on the latest stable v9 the onRest callback  only passes 1 argument. Because of this examples like notification center which rely on this callback for cleanup are no longer possible.

## To Reproduce

- go to this sandbox: https://codesandbox.io/s/ecstatic-bardeen-crw15
- open the console for codesandbox
- go to codesandbox’s dependencies
- switch react-spring from v9 to v9RC1

## Expected behavior

the onRest callback is passed 2 arguments

codesandbox demo: 
https://codesandbox.io/s/ecstatic-bardeen-crw15

## Environment

- `react-spring` v9 (worked on v9RC)
- `react` v17.0.2

joshuaellis:
### Why

resolves #1372
resolves #1041

### What

- `useTransition.config` option now accepts `(item, index, state)` as props (state has been added)
- `onStart`, `onChange`, `onRest` have more similar apis in useTransition with `(value, item, ctrl)` signature (except for the onStart fn which doesn't have a value

### Checklist

- [x] Demo added
- [x] Ready to be merged


dbismut:
As per our discussion, here is what's missing:

- [ ] Fix `onStart` not having a `value` argument in its signature.
  - At the moment `onStart` is called after the first frame, so values in `onStart` are dirty...
  - We should call `onStart` with actual starting values. I see two ways to do this:
    1. move onStart call elsewhere (but not sure where)
    2. have the first frame run without changing values, like a dry run (not sure how but I have the feeling this might be simpler)
   
- [ ] fix `cancel`
- [ ] check that `onChange` / `onStart` / `onRest` in other hooks also run properly and consistently (minus the `item` arg obviously)

joshuaellis:
@dbismut for `onStart` we pass the first argument as 

```js
{
 values: null,
 finished: false,
 cancelled: false
}
```

Just as a halfway for now.

dbismut:
Ok, I'll try to have a look at how to get starting values if that's possible.

PR #1416, opened 2021-04-12
Use tilde range dependencies for stable releases

### LINKED ISSUE #1414: Use exact or ~ deps version in stable releases package.json
## 🤓 Question

As I can read [here](https://github.com/pmndrs/react-spring/blob/master/scripts/prepare.js#L187), we use to put the exact versions of react-spring dependencies in package.json for beta/canary versions but not for stable releases. Is there a particular reason?
The thing is that not fixing the versions in stable releases make them unusable in the future, if a user doesn't want to use the very last stable release and also breaks lockfile maintenance pushed by some dependencies maintaining bots automatically.

### My example
I use the `v9.0.0` version and the lockfile maintenance is breaking because it installs the `v9.1.0` dependencies in my lockfile even if I fixed the `v9.0.0` version of `react-spring` in my package.json (which is normal since this is about sub-dependencies here).

### Proposal
Since you pushed a potential breaking change in `9.1.0` (it seems to break on my side), it shouldn't be pushed automatically to users. But you may want to push patch automatically for security fixes? I would suggest you to use `~` instead of `^` (see [npm docs](https://github.com/npm/node-semver#tilde-ranges-123-12-1)). 
In that case it would push automatically versions like `9.0.1` or `9.0.5` because there are just patches but wouldn't for `9.1.0` since there is a minor risk of breaking changes.

What do you think? I can help to do a PR if needed (is it only related to `prepare.js` file?)

### Current workaround
The workaround to fix this on the user side, which is clearly not ideal, is to pin the versions in the package.json resolutions like:
```
 "resolutions": {
    "@react-spring/animated": "9.0.0",
    "@react-spring/core": "9.0.0",
    "@react-spring/konva": "9.0.0",
    "@react-spring/native": "9.0.0",
    "@react-spring/shared": "9.0.0",
    "@react-spring/three": "9.0.0",
    "@react-spring/web": "9.0.0",
    "@react-spring/zdog": "9.0.0"
}
```

CPatchane:
### Why

See #1414 

### What

Fix the prepare JS script

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

Fix #1414

### DIFF
--- a/scripts/prepare.js
+++ b/scripts/prepare.js
@@ -185,7 +185,7 @@ async function prepare() {
     })
 
     // Replace "link:" versions with (A) exact versions for canary and beta releases,
-    // or with (B) caret ranges for actual releases.
+    // or with (B) tilde ranges for actual releases (allow to get patches automatically).
     const exactRE = /-(rc|canary|beta)\./
     const deps = pkg.dependencies
     if (deps) {
@@ -197,7 +197,7 @@ async function prepare() {
         const dep = packages[depId]
         if (dep) {
           const { version } = packages[depId]
-          deps[depId] = (exactRE.test(version) ? '' : '^') + version
+          deps[depId] = (exactRE.test(version) ? '' : '~') + version
           if (localId !== depId) {
             delete deps[localId]
           }


joshuaellis (approved):
Thanks!

PR #1444, opened 2021-04-22
Fix: Decay (a.k.a intertia) with immediate

### LINKED ISSUE #1026: Inertia/momentum incorrect behavior, and unexpected value jumping
## 🐛 Bug Report

This is a bug for tracking the overall not-quite-right behavior of 1:1 animation with momentum at the end. I started thinking about this after my [issue on react-use-gesture last weekend](https://github.com/react-spring/react-use-gesture/issues/173). Here are some earlier reports and discussions from react-spring and react-use-gesture:

https://github.com/react-spring/react-spring/issues/887
https://github.com/react-spring/react-spring/issues/1010
https://github.com/react-spring/react-use-gesture/issues/132

This is a problem that comes up most commonly with gestures, but I believe the solution lies in the animation primitives. Mapping gestures 1:1 with the animation requires using the `immediate` parameter, but doing so ends up breaking the expected velocity/momentum for the end of the animation/gesture. You can get kind of close to the desired behavior by using a stiff spring config without the `immediate` parameter, but this isn't ideal. HCI research has shown that the desired behavior for these types of gestures is a 1:1 correspondence.

The previous suggested solutions for building this behavior have been custom hooks outside the library, but hopefully we can all agree that this is something that should be covered by the library. Previous solutions added a concept of `decay`, but I think the library already has all of the concepts/APIs we need, it's just a matter of fixing how velocity is calculated when using `immediate`.

## Expected behavior with reproduction

This is a demo of some horizontal scroll areas nested in a vertical scrolling area. Both of them have "touch" scrolling enabled; the `scrollLeft` and `scrollTop` are being animated, in order to co-exist with native browser scrolling.

https://codesandbox.io/s/react-spring-inertial-scroll-challenge-qh4wr

The desired behavior is exemplified by the native scrolling behavior on macOS with a trackpad -- 1:1 mapping of scroll gesture to position, with momentum proportional to the spring config when ending the gesture with velocity. The animated value should not overshoot its target.

The touch scrolling demo implementation does have some momentum, but only a tiny bit; not at all like you'd expect. Here's a gif showing how much momentum I could get when trying my hardest to "sling" it

missing inertia

## Environment

- `react-spring` v9.0.0-rc.3
- `react` v16.8.12

### LINKED ISSUE #1437: Decay animation jumps at the last frame
## 🐛 Bug Report

When a decay animation ends, in the last frame the value resets to "to". The decay animation isn't supposed to use "to" though.
Reference: https://github.com/pmndrs/react-spring/issues/887#issuecomment-823856024
Probably happens because of [this](https://github.com/pmndrs/react-spring/blob/f2170a39111debbca52d0cf233e2f7c43a851b77/packages/core/src/SpringValue.ts#L340) (no decay animation check).

## To Reproduce

[This](https://codesandbox.io/s/decay-v901-jump-cvwgi) is a sandbox that replicates the issue.
simplescreenrecorder-2021-04-21_14 15 22

## Expected behavior

No jump at the end of the decay animation

https://codesandbox.io/s/decay-v901-jump-cvwgi

## Environment

- `react-spring` v9.0.1

joshuaellis:
### Why

This PR should resolve #1026 & because it's bugging me at the same time it should resolve #1437 

### What

If an animation has `immediate` set and that value changes it does not look to "start" the animation at all, it will if you change a value such as `decay` so the following code would work:

```js
if(down)}{
 api.start({ pos, immediate: true })
}
else {
 api.start({ pos, config: { velocity: 2, decay: true } })
}
```

but in reality this is cumbersome and the following should work:

```js
api.start({
 pos,
 immediate: down,
 config: { velocity 2, decay: true }
})
```

Which it now does.

### Checklist

- [ ] Documentation updated
- [x] Demo added
- [ ] Ready to be merged


HQ92:
Looking forward to this fix as I think it's the last piece of the puzzle to making natural feeling touch friendly draggable sliders with real physics based animations without any weird jumping! Looking forward to testing it out.

You can see the issue in my draggable slider currently at https://codesandbox.io/s/react-spring-912-inertial-spring-kyq3d?file=/src/index.js, I'm happy to test once it's been pushed.

PR #1455, opened 2021-04-23
fix(parallax): scrollTo gets correct scroll value

### LINKED ISSUE #1454: Parallax: scrollTo not executing correctly
## 🐛 Bug Report

After playing with the sandbox demos last night, I noticed that the `scrollTo` function in parallax only ever scrolled to `0` and that it wasn't animated. After digging a little bit I discovered that the value of `scroll` in the controller was being passed into `onChange` incorrectly(?) [here](https://github.com/pmndrs/react-spring/blob/4d07bb45dc970b7968c759f0e32ac27075fe703e/packages/parallax/src/index.tsx#L214).

Currently it looks like this:
```javascript
onChange({ scroll }: any) {
  container[scrollType] = scroll
}
```
I got it to work by changing it to this:

```javascript
onChange({ value: {scroll} }: any) {
  container[scrollType] = scroll
}
```
I don't know anything about the new API, but some liberally placed `console.log`s revealed that in the current version, `scroll` always equals `undefined` in `onChange`.

I have a PR in the works, but wanted to open an issue first for better documentation history.
## To Reproduce

* Attach a click listener to any content nested in Parallax and pass `scrollTo` in and click.
* If Parallax is at `scrollTop === 0`, nothing will happen, even if the offset passed to the function is positive.
* If Parallax is at `scrollTop > 0` then Parallax will jump to the top without animating, no matter what offset is passed.

## Expected behavior

Parallax should perform an animated scroll to the entered offset.

[sandbox demo: parallax vert](https://codesandbox.io/s/github/pmndrs/react-spring/tree/master/demo/src/sandboxes/parallax-vert)

## Environment

- `react-spring` v9.1.2
- `react` v17.0.1

kindoflew:
### Why
Resolves #1454. Parallax will now animate `scrollTo` correctly.

### What
Changed the destructuring of the `value` of  `controller` in `onChange`.

I also added a line at 212 to set the `controller`'s value to `state.current` before animating because otherwise it started with stale state (In the current version `state.current` will cause the `controller` to always start at the number from the previous call to `scrollTo`).

### Checklist

- [ ] Documentation updated: n/a
- [ ] Demo added: n/a
- [x] Ready to be merged

### DIFF
--- a/packages/parallax/src/index.tsx
+++ b/packages/parallax/src/index.tsx
@@ -208,10 +208,12 @@ export const Parallax = React.memo(
       const scrollType = getScrollType(horizontal)
 
       state.offset = offset
+
+      state.controller.set({ scroll: state.current })
       state.controller.stop().start({
         scroll: offset * state.space,
         config,
-        onChange({ scroll }: any) {
+        onChange({ value: { scroll } }: any) {
           container[scrollType] = scroll
         },
       })


joshuaellis:
Would me looking at one of the demos be a good way of testing this change? If so which one?

kindoflew:
Yes, the vertical one. The horizontal one still looks weird (PR for that is next).

joshuaellis (approved):
Great fix, thank you!

PR #1458, opened 2021-04-24
fix(parallax): child horizontal prop defaults to parent's

kindoflew:
### Why
Currently, when `Parallax` has `horizontal={true}`, all `ParallaxLayer`s still translate vertically unless `horizontal={true}` is manually set on all of them. This change makes it so `ParallaxLayer`s' default `horizontal` matches `Parallax`'s.

### What
* Added `horizontal` to `IParallax` interface and `Parallax`'s `state`
* Added a check in `ParallaxLayer` to see if `horizontal === undefined` and reassign it if it is.
* Removed a CSS rule in sandbox demo that was causing numbers to appear too far to the left.

### Checklist

- [ ] Documentation updated: (Pending this being merged)
- [ ] Demo added: n/a
- [x] Ready to be merged

You can test this by looking at the sandbox demo "Parallax" (not vert). Currently, when you scroll horizontally, the layers still translate vertically.

This might be a breaking change if anyone is relying on this behavior, but it wouldn't affect anyone who has already manually set `horizontal` on `ParallaxLayer`s.

I will also submit a PR for the docs to add `horizontal` to `ParallaxLayer`'s prop list, but will wait until a decision is reached on this so I can accurately describe its behavior. I understand everyone is probably busy with v9 bug fixes, so if this is not a high priority right now, I'll just update the docs to reflect how it currently works.

### DIFF
--- a/demo/src/sandboxes/parallax/src/styles.module.css
+++ b/demo/src/sandboxes/parallax/src/styles.module.css
@@ -20,7 +20,6 @@
 .number span {
   display: inline-block;
   position: relative;
-  transform: translate3d(-35%, 0, 0);
 }
 
 .slopeBegin {
--- a/packages/parallax/src/index.tsx
+++ b/packages/parallax/src/index.tsx
@@ -54,9 +54,6 @@ export const ParallaxLayer = React.memo(
       // Our parent controls our height and position.
       const parent = useContext<IParallax>(ParentContext)
 
-      // Layer's horizontal defaults to parent's horizontal if not set.
-      if (horizontal === undefined) horizontal = parent.horizontal
-
       // This is how we animate.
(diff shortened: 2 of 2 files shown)


kindoflew:
I decided reassigning a prop was probably a smell, so I refactored it to be a new `const` that replaces `horizontal` conditionals in `ParallaxLayer`. All other changes are the same.

joshuaellis (approved):
Thanks for the contribution!! ⭐ I had thought the demo looked odd when I added it...

joshuaellis:
Regarding your PR notes, i think this makes perfect sense imo, as long as we clearly state this is in the docs it'll be fine, i'll also update the changelogs to reflect this `POTENTIALLY_BREAKING_CHANGE`.

Issue #1452, opened 2021-04-23
CSS Filter Grayscale Bugs


amadeus:
## 🐛 Bug Report
Found a few bugs related to react-spring and animating css filters. I've built a small codesandbox that shows the issue.

## To Reproduce
* Mount a component with an animated CSS filter for grayscale, but start it at `100%`
* Both `useSpring` and the `Spring` component will initialize with no grayscale filter (the item should be fully greyscale at a value of `100%`)
* If you perform an animation to have it animate that property a couple additional things will happen, depending on whether you are using `useSpring` or `Spring`.
  * `useSpring` will not animate the prop at all - it will just toggle to the final value around when the animation should finish
  * `Spring` will be slightly glitched on the first animation since it ends up trying to start from `100%`, but after that first animation it should work fine

## Expected behavior
Animating `grayscale` should both animate and initialize at the proper value on both `useSpring` and `Spring`

## Link to repro (highly encouraged)
https://codesandbox.io/s/xenodochial-water-3otuy?file=/src/App.tsx

## Environment

- `react-spring` v9.1.1
- `react` v17

joshuaellis:
Yeah, it looks like the filter is being changed to something odd, if you do:

```js
const { filter } = useSpring({
    config: SPRING_CONFIG,
    filter: `grayscale(${100 - value * 100}%)`,
    backgroundColor: value === 0 ? "red" : "blue"
  })

console.log(filter.get())
```
you see `rgba(128, 128, 128, 1)scale(100%)`... If you're after a quick fix, I encourage you too look at creating a PR.

PR #1460, opened 2021-04-25
Fix: Ignore partial color names when interpolating strings.

CodyJasonBennett:
### Why?

This PR prevents strings like `grayscale` from being interpolated as `gray` and `scale` (color & range). This is seen in #1452.

Here's their sandbox but with this fix: https://codesandbox.io/s/grayscale-demo-react-spring-mq1wl.

### What?

I updated the regex in the shared string interpolator to ignore color names that are immediately followed by alphanumerical characters.

Previously, I had this also check preceding characters, but I couldn't find a use-case for this, and the regex for that [isn't well supported](https://caniuse.com/js-regexp-lookbehind).

### Checklist

- ~~Documentation updated~~
- ~~Demo added~~
- [x] Ready to be merged

### DIFF
--- a/packages/shared/src/stringInterpolation.ts
+++ b/packages/shared/src/stringInterpolation.ts
@@ -36,8 +36,10 @@ export const createStringInterpolator = (
 ) => {
   if (!namedColorRegex)
     namedColorRegex = G.colors
-      ? new RegExp(`(${Object.keys(G.colors).join('|')})`, 'g')
-      : /^\b$/ // never match
+      ? // match color names, ignore partial matches
+        new RegExp(`(?<!\\w)(${Object.keys(G.colors).join('|')})(?!\\w)`, 'g')
+      : // never match
+        /^\b$/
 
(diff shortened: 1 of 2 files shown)


packages/shared/src/stringInterpolation.ts
```diff
@@ -36,8 +36,10 @@ export const createStringInterpolator = (
 ) => {
   if (!namedColorRegex)
     namedColorRegex = G.colors
-      ? new RegExp(`(${Object.keys(G.colors).join('|')})`, 'g')
-      : /^\b$/ // never match
+      ? // match color names, ignore partial matches
+        new RegExp(`(?<!\\w)(${Object.keys(G.colors).join('|')})(?!\\w)`, 'g')
```
CodyJasonBennett, line 40: To clarify, the first group in the regex `(?<!\\w)` is what I mention in regressions.

joshuaellis (approved):
Thanks! ⭐ and you wrote a test 😍

Issue #1049, opened 2020-06-09
Accept zero as valid for transforms


arackaf:
## 🚀 Feature Proposal

This is super minor, and I'll completely understand if you just close it as "won't implement" but I bumped into it so I figured it was at least worth bringing up.

```js
  const transition = useTransition(booksObj, {
    config: { ...config.default }, //config.molasses,
    from: { opacity: 0, position: "static", transform: "translate3d(0, 0px, 0px)" },
    enter: { opacity: 1, position: "static", transform: "translate3d(0, 0px, 0px)" },
    leave: { opacity: 0, position: "absolute", transform: "translate3d(25%, 0px, 0px)" }
  });
```

The transform above does not work, since the zero's in the X slot have no units, even though zero would normally be considered a dimensionless quantity.

The workaround is trivial of course

```js
  const transition = useTransition(booksObj, {
    config: { ...config.default }, //config.molasses,
    from: { opacity: 0, position: "static", transform: "translate3d(0%, 0px, 0px)" },
    enter: { opacity: 1, position: "static", transform: "translate3d(0%, 0px, 0px)" },
    leave: { opacity: 0, position: "absolute", transform: "translate3d(25%, 0px, 0px)" }
  });
```

But it took me a minute or two to figure this out

## Motivation

See above

## Example

See above

aleclarson:
I think a lot of people trip on this. 

The problem is, the current "string interpolation" logic is generic, so I'm a little uncertain how that could be fixed, but I guess adding another global would be good enough. Then the `@react-spring/web` package could inject its own handling for transform strings.

Btw, you probably don't want to write out the transform string manually, since you can use shorthand style props instead: https://aleclarson.github.io/react-spring/v9/#Shorthand-style-props

arackaf:
Whoa - nice tip on the shorthand form - thanks a ton!

joshuaellis:
released in v9.3.0

Issue #643, opened 2019-04-21
Using percentages on width transition in multistage useTransition throws type error


meisterpeeps:
## 🐛 Bug Report

Using percentages on width transition in multistage hook useTransition throws type error
`TypeError: entry.interpolation.calc is not a function`
>   1249 | if (reset || hasNewGoal && currentValueDiffersFromGoal || hasNewConfig) {
>   1250 |   // Convert regular values into animated values, ALWAYS re-use if possible
>   1251 |   if (isNumber || isString) parent = interpolation$$1 = entry.parent || new AnimatedValue(fromValue);else if (isArray) parent = interpolation$$1 = entry.parent || new AnimatedValueArray(fromValue);else if (isInterpolation) {
> > `1252 |     var prev = entry.interpolation && entry.interpolation.calc(entry.parent.value);`
>        | ^  1253 |     prev = prev !== void 0 && !reset ? prev : fromValue;
>   1254 | 
>   1255 |     if (entry.parent) {

## To Reproduce

Use a string percentage in a multi-stage useTransition hook width.
`enter: {width: '20%', opacity: 1}` works
`enter: [{width: '20%'}, {opacity: 1}]` throws error

## Environment

- `react-spring` v8.0.19
- `react` v16.8.2

I also just checked the behavior on the 9 beta version.  The problem still exists, but instead of a type error I get a  

> Warning: `NaN` is an invalid value for the `width` css style property.

-- and the width doesn't render.   On 9 the screen does in fact render though, just with broken css.  On 8 I get a red screen.

aleclarson:
Could you provide a CodeSandbox using `react-spring@next`?

meisterpeeps:
Sure;  Let me hook one up.

meisterpeeps:
Not the most interesting example... but it still works, yah?
https://codesandbox.io/s/yk59609241
The problem occurs also with height.  I just forked one of the examples and updated it to next, and changed the height to a string percentage.

aleclarson:
Using `"0%"` in place of `0` for `height` fixes that example. It seems that interpolating from a number to a percentage is not yet supported.

meisterpeeps:
Awesome.   Since I can go from a percentage to a percentage in my particular use case this actually fixed my problem 😛.   I appreciate the very quick feedback.

aleclarson:
No problem! Let's keep this open as a feature request for number-to-percentage animations (and vice versa).

PR #1473, opened 2021-04-28
fix: interpolate bare numbers with units

CodyJasonBennett:
### Why

Bare numbers are not supported when interpolating to units, as seen in #643 and #1049.

### What

This pull request adds units to bare numbers when interpolating. This does not change or convert between different units.

TODO:
 - [x] Recreate issues in sandbox & verify fix propagates to hooks

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


CodyJasonBennett:
Should I point this at master instead?

joshuaellis:
Keep at `beta` for now so we dont loose track of which is in sync, i'll probably merge beta to master and delete beta over the next couple of days, I've just been double checking the build stuff.

joshuaellis:
Also, i've just realised what you're looking at, awesome 💯 in-fact I can hold off on 9.2.0 until this is in, that'd be cool (there's no rush, we can keep releasing betas with fixes in the meantime).

joshuaellis:
Is this PR ready in terms of functionality? We just need to test that it fixes the sandbox issues or is there more?

CodyJasonBennett:
Yes, but I'll have to make sure that the fix fully propagates to the hooks.

This fixes the shared interpolation helpers.

CodyJasonBennett:
Oh no, I didn't see this was still pending.

Although it doesn't look like anything broke since being implemented, this should be confined to web.

joshuaellis:
Why do you think it needs to be confined to `web`?

CodyJasonBennett:
I see. I was over-scoping the PR to validate a bit more than what's needed.

Issue #1453, opened 2021-04-23
useChain does not work as expected for transitions


martinkutter:
## 🐛 Bug Report

`useChain` does not chain animations correct, if `Transition` should animate after `Spring`.

## To Reproduce
`Spring` first, than `Transition` -> does not work
See https://codesandbox.io/s/wizardly-leaf-b4v5g?file=/src/App.tsx

`Transition` first, than `Spring` -> does work
comment out line 38; comment in line 40
## Expected behavior
`Transition` animation should run after `Spring` animation, not in sequence.

## Environment

- `react-spring` v9.1.2
- `react` v17.0.1

martinkutter:
What I have found out so far: 
I think the issue is, that there are no controllers available for transitions in https://github.com/pmndrs/react-spring/blob/master/packages/core/src/hooks/useChain.ts#L15

joshuaellis:
Firstly, I think you need to set an actual height on the inner `divs` the don't appear and it's confusing.
Secondly, and this is more so i understand, you've removed the data from the transition, but you're expecting the divs to stay there until the spring has changed background color and then the inner divs animate out?

martinkutter:
Ups... :( I had forgotten to save the CSS file.
I think the Example is now much clearer.

https://codesandbox.io/s/wizardly-leaf-b4v5g?file=/src/App.tsx

https://user-images.githubusercontent.com/16589094/116391969-24844480-a820-11eb-84de-b9d48906838d.mp4
In the first part of the screen capture the animation starts parallel not chained. In the second part the chain starts with the transition and the animations runs chained as expected.

joshuaellis:
So the issue is the background color changes whilst the transitions are animating in. but actually, the background color should change **then** the divs animate in?

martinkutter:
Yes correct!

martinkutter:
I had the chance to take a closer look at the problem with the missing controller.

If I understand correct at this line the `controller` should be assigned to `ref` from `props`.
https://github.com/pmndrs/react-spring/blob/9568020efb64addb97816232759528a7681ed97c/packages/core/src/hooks/useTransition.tsx#L295

But `payload.ref` is not (and cannot be) defined. `payload` is initialized in 
https://github.com/pmndrs/react-spring/blob/9568020efb64addb97816232759528a7681ed97c/packages/core/src/hooks/useTransition.tsx#L206
[...]

PR #1475, opened 2021-04-28
fix: assign controller to provided ref for useTransition

martinkutter:
### Why
Injected `ref` is not updated with controller for `useTransition`, as seen in #1453

### What
This pull request ensures that injected `ref` is updated.

### Checklist
- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


packages/core/src/hooks/useTransition.test.tsx
```diff
@@ -35,6 +36,17 @@ describe('useTransition', () => {
     expect(rendered).toEqual([false])
   })
 
+  it('assign controller to provided "ref"', async () => {
```
joshuaellis, line 39: Can we make a more realistic test in the case of `useTransition` i.e. typically we might have 3 items in our Transition, so in my head, the ref should have three `controller`s?

joshuaellis (changes requested):

packages/core/src/hooks/useTransition.test.tsx
martinkutter: :+1: good idea -> updated

joshuaellis (approved):

joshuaellis:
Thanks for the great contribution ⭐

PR #1480, opened 2021-04-29
[fix] loop bugs

### LINKED ISSUE #1063: What is the proper way the Loop a Trail in v9?
## 🤓 Question

What is the proper way to loop a Trail in v9.0? I've had success with the loop prop for Springs but I'm having difficulty using it with a Trail.

```js
const paths = [...]

<Trail
   loop
   items={paths}
   from={{ opacity: 0.2, y: 0, stroke: 'rgba(255,75,75,1)'}}
   to={{ opacity: 1, y: 30, stroke: 'rgba(80,80,80,.35)'}}
>
   {(item, index) => ({ y, opacity, stroke }) => (
      <animated.path
         style={{
            opacity,
            transform: y.interpolate(y => `translateY(${ y}px`),
         }}
         d={item}
         fill="black"
         stroke={stroke}
         stroke-width={3}
      />
   )}
</Trail>
```
This code runs the full animation once, but on subsequent loops, only the first few elements in the array are animated. I've tried playing around with delay and duration and it seems to improve the animation somewhat but the latter items in the array only partially animate. I also tried initially setting loop to false, then toggling it to true in the onRest function, but this doesn't loop at all.

Here is a link to the sandbox: [https://codesandbox.io/s/tender-https-55nmz?file=/src/LoopedTrail.js](https://codesandbox.io/s/tender-https-55nmz?file=/src/LoopedTrail.js)

joshuaellis:
### Why

Should resolve #1063

### What

*TODO*

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

### DIFF
--- a/.changeset/popular-brooms-wash.md
+++ b/.changeset/popular-brooms-wash.md
@@ -1,0 +1,16 @@
+---
+'@react-spring/animated': patch
+'@react-spring/core': patch
+'@react-spring/parallax': patch
+'react-spring': patch
+'@react-spring/shared': patch
+'@react-spring/types': patch
+'@react-spring/konva': patch
+'@react-spring/native': patch
+'@react-spring/three': patch
+'@react-spring/three-v5': patch
+'@react-spring/web': patch
+'@react-spring/zdog': patch
+---
+
+support react17 peer deps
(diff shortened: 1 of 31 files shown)


joshuaellis:
note: i think there's only one SpringValue shared between the all the Controllers in a trail?

joshuaellis:
I'm closing for now as this bug has me stumped, I might come back to it, or someone else might want to try and fix it.

Issue #619, opened 2019-04-03
Sticky parallax


mogmog:
Many parallax libraries have the ability to mark an element as 'sticky' such that as you scroll, certain elements are not scrolled until a certain threshold is reached, eg:

https://stickyroll.github.io/react-stickyroll/

My question is, does react-spring's parallax provide such a feature, or is there a property i can inspect to simulate this effect?

fullspectrumprocessing:
Definitely interested in this feature also!!

nickmcmillan:
I needed this for a project so decided to try writing it myself. The effect can be accomplished by using React Springs primitives and `position: sticky`. 

I've loaded it with comments. Hopefully some of you find this useful and also spot some areas for improvement.

https://codesandbox.io/s/scroll-animation-with-react-spring-pndnu

joshuaellis:
closing due to inactivity. please use discussions or discord if you want more help or please consider creating a PR if you want to add a feature 😄

joshuaellis:
Big thanks to the fantastic @kindoflew this has been released in the latest 9.2.0 beta (6).

PR #1482, opened 2021-04-30
feat(parallax): sticky ParallaxLayer

kindoflew:
### Why

Rough draft of implementing the feature request from #619. Adds `sticky: number` prop which acts as a second offset where the `ParallaxLayer`will behave as if `position: sticky` was set from `offset` to `sticky`. When outside of those bounds, the layer behaves as if it had `speed={0}` and `offset={offset | sticky}` depending on which side of the bounds you're on. It could also be an object, something like `sticky={{ from: number, to: number }}` and ignore `offset` completely.

### What

This is less straightforward than it seemed like it'd be. We can't actually use `position: sticky` (AFAIK) because the inner `Parallax` container has `overflow: hidden` and [CSS hates fun](https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky).

To get around that, the `setPosition` method checks to see if `sticky` is set and if it is either: 
1) sets `translate` to `scrollTop` or
2) keeps `translate` at either the `offset` or `sticky` coordinate

with `immediate = true` for both cases (I had an issue with sticky layers overshooting and settling with spring physics, even with `speed={0}`/`clamp: true` and this solved it). To avoid all the ternaries and nested conditionals, I'd probably split this into it's own `setSticky` method (or even its own `StickyLayer` component?), but I just wanted to make it work first. However, I did notice occasional jank on my end when the layer is "sticky", but it's entirely possible that I'm using the react-spring functions in an inefficient way, so any comments on that would be welcomed.

 Also, to tidy up the render function, I added  a `state` const to `ParallaxLayer` using `useMemoOne` to hold `horizontal` and `translate`. `state.horizontal` then replaced `scrollHorizontal`. There's also a very crude demo added to the sandbox to test this.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

Eventually, we could also expose progress information, similar to libraries like react-stickyroll. That way users could hook in other react-spring animations to happen in time with certain landmarks. But I'm getting ahead of myself. 

Anyway, if this is wanted it seems our best choice is some kind of translate trickery.


joshuaellis:
I've managed to understand the feature request now, and this is something I regularly get asked to look into implementing for clients, so I do think it's a great feature to include. Maybe this PR could be the start of our Parallax testing?

kindoflew:
Sounds good to me! My current hangup is wondering if this functionality should be added to `ParallaxLayer` or if it should be its own separate `StickyLayer` component. If it stayed as a prop on the current component, all other props would be ignored and I'm not sure if that violates the principal of least surprise (we could just make sure it's very clear in the docs).  

On the other hand, if we split them, there would probably be a decent amount of duplicate code between the two layer components.

In addition to all that, I'm still trying to figure out a way to actually use `position: sticky` which would make everything much easier and (probably) more performant regardless of which version we go with.

kindoflew:
I basically have everything set up and working, but I'm having some trouble figuring out exactly what `config` to use when animating `translate` if the layer is "sticky". I currently have it set to `{ duration: 0 }`, which works fine most of the time, but I get occasional jank if scrolling quickly. I think it's because the value isn't being interpolated between each `scrollTop` integer so it'll jump a little bit on any frame that should be between two whole numbers. I've tried some other animated configs, but I can't figure out how to have instant updates and still keep interpolation.

joshuaellis:
> Sounds good to me! My current hangup is wondering if this functionality should be added to ParallaxLayer or if it should be its own separate StickyLayer component. If it stayed as a prop on the current component, all other props would be ignored and I'm not sure if that violates the principal of least surprise (we could just make sure it's very clear in the docs).

I think for the time being it could be the same component (DRY). But lets see when it's finished if it makes sense...

>  I've tried some other animated configs, but I can't figure out how to have instant updates and still keep interpolation.

I'll try find sometime tomorrow to have a look at this and hopefully have an informed opinion ⭐

kindoflew:
I think I figured it out! If the inner `Parallax` container is set to `overflow: clip` we can use `position: sticky` on a layer. `clip` works almost exactly like `hidden` except that it disallows scrolling, but that doesn't matter in our case because the *outer* container is the one that is being scrolled. The other two parallax demos still work with this change (but I'm sure we'll feel more secure once we write these tests :sweat_smile:). Working out the details now, I'll keep you updated.

kindoflew:
Changed the following:
* change `overflow: hidden` to `overflow: clip` in `Parallax`'s inner content wrapper
* add a `useRef` to `ParallaxLayer`
* refactor sticky position calculation into its own function, `setSticky`, which is conditionally called from `setPosition`
* add an `isSticky: boolean` property to `layer`
* add conditional `top`/`left` rule to `ParallaxLayer`'s `style` (has no effect on normally translating layers)

I tested the other two demos manually and they still work as expected. 

With the way it's currently designed, if `sticky` is set, both `sticky.from` and `sticky.to` would be required props.

kindoflew:
I think I'm pretty happy with the internal implementation, all that's left is to solidify the API. As of right now it's as follows:
* `sticky` is a prop of type `{ from: number, to: number }`
* when passing the `sticky` prop to `ParallaxLayer`, both `from` and `to` are required*
* when `sticky` is set, all other props are ignored
* the layer will be sticky between the `from` and `to` offsets and will behave as if it has `rate={0}` and `offset={ from | to }` otherwise (not using spring physics when not sticky is to avoid awkward snapping into position when transitioning between sticky/absolute)

\* it might be cool to have a default config for `sticky` so you could do things like this:
```jsx
<ParallaxLayer sticky={{ from: 1 }}>
// or
<ParallaxLayer sticky={{ to: 3 }}>
```
Or maybe have it be a union type of `{ from: number, to: number } | boolean` so you could do:
```jsx
<ParallaxLayer sticky>
```
And it would always be sticky.

I don't know if either of those are necessary, but it might make for a more flexible DX.

kindoflew:
So, I might have jumped the gun. It turns out that `overflow: clip` is new to the spec and Safari doesn't support it at all and Chrome only started in its most recent version. I've come up with another solution, but it involves moving the layer *outside* of the inner container and into the outer container (becoming a sibling of the inner). This can be achieved two ways. First like this in `ParallaxLayer`:
```javascript
useOnce(() => {
  if (sticky) {
    const layer = layerRef.current
    const container = layer.parentNode.parentNode
    container.appendChild(layer)
  }
})
```
Or in `Parallax` by mapping `React.Children` and rendering non-sticky children in the inner container, and sticky children outside of it. I'm not sure which one is the more "React" way, but they both seem to work. This doesn't affect the other demos.

joshuaellis:
don't do the `useOnce` method 😅 do the mapping

joshuaellis:
Sorry, i was really busy today and i wanted to put that before you spent time doing the `useOnce` method! Yeah, I wouldn't append things to the dom like that. Using the mapping is the "correct" way in react imo. This sort of thing can lead to unintended side effects e.g. the node just disappearing.

kindoflew:
No worries! Ah, that makes sense, thanks.

The sticky functionality still works the same, but I feel like I have reservations about doing it this way and I can't put my finger on why. It *shouldn't* affect anything else because everything is absolutely positioned and outside of the normal document flow.

One thing I noticed about this change is that it causes the z-index of a sticky layer to always be higher than any non-sticky layer, but that can be adjusted manually by the user.

kindoflew:
If we're committed (for now) to keeping it as one component I think it's in a pretty good place for review. But the more I look at it, all the `if (sticky)` checks everywhere kind of hints to me that maybe it should be separate? How it is now, almost every part of `ParallaxLayer` has different behavior depending on whether it's sticky or not. On the other hand, it's not a very complex component, so I don't think it makes it hard to tell what's going on.

packages/parallax/src/index.tsx
```diff
@@ -274,22 +309,33 @@ export const Parallax = React.memo(
...
+                position: 'absolute',
+                [horizontal ? 'height' : 'width']: '100%',
+                [horizontal ? 'width' : 'height']: state.space * pages,
+                WebkitTransform: START_TRANSLATE,
+                msTransform: START_TRANSLATE,
+                transform: START_TRANSLATE_3D,
+                ...props.innerStyle,
+              }}>
+              <ParentContext.Provider value={state}>
+                {React.Children.map(
```
joshuaellis, line 326: Couldn't this just be `rest.children.map(child => child.props.sticky && child)`?

packages/parallax/src/index.tsx
```diff
@@ -46,40 +48,58 @@ export interface ParallaxLayerProps extends ViewProps {
...
   speed?: number
+  /** Layer will be sticky between these two offsets, all other props are ignored */
+  sticky?: { from: number; to: number }
 }
 
 export const ParallaxLayer = React.memo(
   React.forwardRef<IParallaxLayer, ParallaxLayerProps>(
-    ({ horizontal, factor = 1, offset = 0, speed = 0, ...rest }, ref) => {
+    (
+      { horizontal, factor = 1, offset = 0, speed = 0, sticky, ...rest },
```
joshuaellis, line 58: Is there a reason we're not destructing things like `children` when we're calling them from before?

packages/parallax/src/index.tsx
```diff
@@ -46,40 +48,58 @@ export interface ParallaxLayerProps extends ViewProps {
...
       // Our parent controls our height and position.
       const parent = useContext<IParallax>(ParentContext)
 
       // This is how we animate.
       const ctrl = useMemoOne(() => {
-        const targetScroll = Math.floor(offset) * parent.space
-        const distance = parent.space * offset + targetScroll * speed
+        let translate
+        if (sticky) {
+          translate = sticky.from * parent.space
```
joshuaellis, line 68: do we need a default for `sticky.from` how does a controller normally handle is if there no `from` defined but a `to` prop?

packages/parallax/src/index.tsx
```diff
@@ -90,6 +110,23 @@ export const ParallaxLayer = React.memo(
 
       React.useImperativeHandle(ref, () => layer)
 
+      const layerRef = useRef<any>()
+
+      const setSticky = (height: number, scrollTop: number) => {
+        const from = sticky!.from * height
```
joshuaellis, line 116: This is a similar thing to above, we should look at how `Controller` handles a lack of `from` or `to`

joshuaellis (changes requested):
Some thoughts for discussion, but great work regardless ⭐

packages/parallax/src/index.tsx
kindoflew: I thought so too, but I was getting a TypeScript error: `Property 'map' does not exist on type string` and so I thought `React.Children` was the way to go about it (also, it deals with the edge case of having only one `ParallaxLayer`, in which case mapping `children` would fail because it's not an array). I did manage to get it down to this just now (after destructing):
```
React.Children.map(children, (child: any) => child.props.sticky && child)
```
`children.map...` also works in this case, but will still fail to render anything if there is only one child. Though, to be fair, I can't see a lot of situations where there would only be one.

joshuaellis: Ah yes, one child, although you're right is a bit of a weird case.... Lets keep it with what you have ⭐  it was a tiny point anyway 👍🏼

packages/parallax/src/index.tsx
kindoflew: My initial thoughts are: `sticky.from` should default to `0` and if no `sticky.to` is set, the layer should just always be sticky whenever the container is scrolled above `sticky.from`. Ideally `sticky.to` would default to `parent.pages * parent.space`, but then we'd have to include `pages` in `ParentContext`. 

I don't know if the behavior really mirrors `from`/`to` of `Controller` and similar APIs, so maybe it should be changed to something like `start`/`end`?

joshuaellis: `0` makes sense to me. Maybe `to` could be `1`? or `1` greater than `from`?

Yeah `start` and `end` makes more sense to me I think, is that weird in comparison to the other `parallax` props?

kindoflew: Ooh, I like `to = from + 1`! 

As far as the names, I don't think it'd be too weird. Either choice will sound a little off I think because most `from`/`to` props in react-spring are referring to interpolation of some sort and this is about offsets. I only picked `from`/`to` initially for consistency, but I also liked how it sounded out loud, "this layer is sticky `from` [here] `to` [here]."

kindoflew:
Alright, I cleaned up the child mapping a little and changed the `sticky` props to `start` and `end`. I also added the default values. Did you still want to start testing in this PR?

joshuaellis:
No I think we should make a new PR actually 😅  my brain vastly underestimate the work required lol

kindoflew:
Right? I figured I'd just slap `position: sticky` somewhere and call it a day. 

Anything else I need to do before this is ready to merge? Does the demo look ok?

joshuaellis:
Ah so good 🤩

joshuaellis (approved):
This is fantastic, honestly 🙇🏼‍♀️

kindoflew:
Thanks so much! :blush:

PR #1493, opened 2021-05-08
[fix] useTransition ref handling

### LINKED ISSUE #1453: useChain does not work as expected for transitions

joshuaellis:
### Why

`useTransition` was not passing its ref as part of the payload, this the controllers (each item has a controller) were not being attached. This caused failures in instances such as useChain. 

This resolves #1453 

### What

Pass the ref as part of the payload. However, when passing a ref, transitions will not be called unless `ref.start` is called, this is important and an example of what this looks like can be seen here: https://codesandbox.io/s/goofy-allen-u4hml?file=/src/App.tsx

I've also added the same API pattern that `useSprings` has, where only supplying a function as a prop argument will return a ref like so:

```js
const [transition, transApi] = useTransition(items, () => ({ ...props }))
```

### Checklist

- [ ] Documentation updated – we're going on to beta so won't do yet.
- [x] Demo added
- [x] Ready to be merged

### DIFF
--- a/demo/src/sandboxes/simple-transition/src/App.tsx
+++ b/demo/src/sandboxes/simple-transition/src/App.tsx
@@ -1,5 +1,5 @@
-import React, { useState, useCallback, CSSProperties } from 'react'
-import { useTransition, animated, AnimatedProps } from '@react-spring/web'
+import React, { useState, useCallback, CSSProperties, useEffect } from 'react'
+import { useTransition, animated, AnimatedProps, useSpringRef } from '@react-spring/web'
 
 import styles from './styles.module.css'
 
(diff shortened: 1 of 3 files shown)


joshuaellis:
@dbismut See the sandbox above, do you think it's weird that you even if you change the transition data, you need to call `start` to trigger said animation? I find it odd to begin with when thinking in isolation of the `useTransition` hook but as a whole compared to our other hooks, it's consistent which I prefer. Good to hear your thoughts 🤔

PR #1496, opened 2021-05-09
[fix] Delay conflict with Trail in useTransition

### LINKED ISSUE #1099: Delay and trail conflict on useTransition (v9.0.0-rc3)
Hi there! I ran into an issue while creating a multi-stage transition using react-spring.

## 🐛 Bug Report
As soon as a delay is placed as a property somewhere in the transition, any trail is immediately disregarded.

## To Reproduce
See this CodeSandbox: [https://codesandbox.io/s/dank-leftpad-hyvle?file=/src/App.js](https://codesandbox.io/s/dank-leftpad-hyvle?file=/src/App.js)
If a delay is present on any of `from`, `enter` or `leave`, any value of trail is disregarded. As soon as all mentions of delay are removed, the trail shows up as normal.

## Expected behavior
I would expect that a delay and trail could be combined, where a delay precedes the staggered set of items following the trail. The typescript annotations show that both trail and delay can be part of the global config, as well as the individual stages, so I figured this should have been a valid combination. If it is not, that is okay too.

[https://codesandbox.io/s/dank-leftpad-hyvle?file=/src/App.js](https://codesandbox.io/s/dank-leftpad-hyvle?file=/src/App.js)

## Environment

- `react-spring` v9.0.0-rc3
- `react` v16.12.0

Thanks for making something awesome!

joshuaellis:
### Why

resolves #1099 

### What

Since we handle the trail prop in `useTransition` and pass that as our `delay` prop to the `Controller` we therefore need to call the original `delay` prop from the props passed and add that to the `payload` to ensure timings are correct.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useTransition.tsx
+++ b/packages/core/src/hooks/useTransition.tsx
@@ -168,6 +168,7 @@ export function useTransition(
 
     let to: TransitionTo<any>
     let phase: TransitionPhase
+    let propsDelay = callProp(props.delay || 0, key)
     if (prevPhase == TransitionPhase.MOUNT) {
       to = props.enter
       phase = TransitionPhase.ENTER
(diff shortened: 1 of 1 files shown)


joshuaellis:
@leinelissen it'd be great if you could confirm this works as expected? Check this slightly modified sandbox out, it uses the PR as `react-spring` – https://codesandbox.io/s/youthful-boyd-8u981?file=/src/App.js

leinelissen:
@joshuaellis Looks like how I was expecting things to work, thanks so much for looking at this. As per your phase styles comment: I found that `delay` and `trail` were part of the Typescript bindings for the phase styles object, and as such I was confused on whether this was an accepted setup. I am completely fine with it being static for the transition.

joshuaellis:
Yeah I did find that odd, I think it's because at the route of `phases` it's just our `to` object with I *believe* is based on a `[key:string]: any` format to allow loads of flexibility... Before I merge this I might check out we have a "not allowed" list of keywords 🤔

But i'm glad this has solved your issue.

joshuaellis:
No i've checked, because its to just CSS props you might wanna use, we let most things go through, I guess thats why documentation is important 😅

leinelissen:
Right! I've seeing ghosts then 😅 Thanks again for looking at this.

PR #1501, opened 2021-05-10
fix: csp issue in SpringRef

### LINKED ISSUE #1423: CSP issue
## 🐛 Bug Report

As of the [this commit](https://github.com/pmndrs/react-spring/commit/f3c4c0b7fbc9db4112e3722f3c05269b73b49b08#diff-bd9741bffd771a2193b84cd7317ae87f02c9d811e9b6aaba5d634a0e16a1fe1aR23-R25) React Spring won't work whenever a CSP without `'unsafe-eval'` directive is active.

## To Reproduce

Set content security policy (CSP) without `'unsafe-eval'` in `connect-src` directive.

joshuaellis:
### Why

resolves #1423

### What

Replace the string call with a `Proxy`. This isn't supported by ie11, but officially we don't support it.
There could be performance downsides so we need to test.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/packages/core/src/SpringRef.ts
+++ b/packages/core/src/SpringRef.ts
@@ -21,7 +21,10 @@ export class SpringRef<State extends Lookup = Lookup> extends Function {
   readonly current: Controller<State>[] = []
 
   constructor() {
-    super('return arguments.callee._call.apply(arguments.callee, arguments)')
+    super()
+    return new Proxy(this, {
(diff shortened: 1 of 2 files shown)


joshuaellis:
Leaving this here so I don't loose it: https://codesandbox.io/s/spring-perf-w-proxy-0n526?file=/src/App.tsx

PR #1503, opened 2021-05-11
[Fix] csp issue

### LINKED ISSUE #1423: CSP issue

joshuaellis:
### Why

To resolve #1423. In short the current way we're allowing people to call the `SpringRef` as they used to in v8 is to extend the `SpringRef` class from `Function` this creates issues with CSPs. Therefore we're at a fork in the road of implementing this PR, or simply removing the functionality earlier than expected.

### What

Currently i've created a wrap-around function named `CallbableSpringRef`. It should allow you to do:

```js
const [styles, start] = useSpring(() => ({
 from: { width: 0 },
 to: { width: 100 }
}))

start()
```

This keeps in line with a feature we brought back in `9.1.0` to ease people into the new v9 [imperative API](https://react-spring.io/common/imperatives-and-refs#api-methods). However this will be deprecated in the future.

It's a bit messy with the TS but I really do see this a bandaid to be removed sooner rather than later.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/demo/src/sandboxes/card/src/App.tsx
+++ b/demo/src/sandboxes/card/src/App.tsx
@@ -44,17 +44,17 @@ export default function App() {
   useGesture(
     {
       onDrag: ({ active, offset: [x, y] }) =>
-        api.start({ x, y, rotateX: 0, rotateY: 0, scale: active ? 1 : 1.1 }),
-      onPinch: ({ offset: [d, a] }) => api.start({ zoom: d / 200, rotateZ: a }),
+        api({ x, y, rotateX: 0, rotateY: 0, scale: active ? 1 : 1.1 }),
+      onPinch: ({ offset: [d, a] }) => api({ zoom: d / 200, rotateZ: a }),
       onMove: ({ xy: [px, py], dragging }) =>
         !dragging &&
-        api.start({
+        api({
           rotateX: calcX(py, y.get()),
           rotateY: calcY(px, x.get()),
           scale: 1.1,
         }),
       onHover: ({ hovering }) =>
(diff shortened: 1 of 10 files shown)


joshuaellis:
@dbismut do you have any thoughts how to tidy up the `ts-ignore`s i've left 😓

joshuaellis:
Actually i'm thinking I should just rewrite `SpringRef` to be a function... so you don't call `new SpringRef()` just `SpringRef()`.

It'd only break for folk who use classes if they're using `SpringRef`.... 🤔 just be cleaner.

CodyJasonBennett (commented):
@joshuaellis, I like the idea of converting `SpringRef` to a function to keep it consistent with `CallableSpringRef` and native hooks.

I'm not sure how much we should worry about breakage as it's an internal class, but whether this is a private API would be my concern — I don't see documentation on it to suggest it's not.

joshuaellis:
It's technically not an internal class, if you want to pass a ref to a spring in a hook you use `useSpringRef` however, if you're writing Class components (and some people do...) you have to use the raw `SpringRef`. However I *imagine* the amount of folk who do use it to be quite small...

I'll work on moving it to a function 👍🏼

PR #1530, opened 2021-05-28
fix(parallax): add position properties to layer

kindoflew:
### Why

Found this bug while making the new demo for testing (I didn't even set up `cypress` yet :sweat_smile:):

If more than 2 layers are sticky on the same page, their positions may be suddenly shifted down a page before becoming sticky when scrolled to a second time (depending on their order in the HTML). Description might not be totally clear, but you can see what I mean here: https://codesandbox.io/s/nifty-night-j6xdm?file=/src/App.tsx. 

Obviously, I had to discover this the day *after* 9.2.0 released.

### What

Instead of having conditional position properties like this:
```javascript
[layer.horizontal ? 'left' : 'top']: 0
```
All four are now listed explicitly (`top: 0, bottom: 0`, etc..). This has no effect on regular layers.

In the future when the `inset` property has [more support](https://caniuse.com/?search=inset) we can switch it to that so it looks a little cleaner, but that's a nitpick really. (Also, by that time, `overflow: clip` will hopefully be better supported and we can just use that instead and none of this will matter anyways.)

### Checklist

- [ ] Documentation updated: N/A
- [ ] Demo added: N/A
- [x] Ready to be merged

### DIFF
--- a/packages/parallax/src/index.tsx
+++ b/packages/parallax/src/index.tsx
@@ -164,7 +164,10 @@ export const ParallaxLayer = React.memo(
           ref={layerRef}
           style={{
             position: 'absolute',
-            [layer.horizontal ? 'left' : 'top']: 0,
+            top: 0,
+            bottom: 0,
+            left: 0,
+            right: 0,
             backgroundSize: 'auto',
             backgroundRepeat: 'no-repeat',
             willChange: 'transform',


joshuaellis (approved):
Good spot! I'll release this now. I'm thinking of adding the changesets bot like in `leva` so contributors will be prompted to make a changeset 🤔

PR #1542, opened 2021-06-04
feat(parallax): add e2e tests

### LINKED ISSUE #1474: [tests] E2E tests for Parallax
## 🚀 Feature Proposal

As part as our going endeavour to be confident in our updates to `react-spring`, we're missing tests for our Parallax package. I don't think `jest` will suffice on this, although i'm happy to be told i'm wrong, I was thinking E2E tests with `cypress` namely using `@testing-library/cypress` because I like the principals of `testing-library`.

## Motivation

We need more confidence when making changes to the `parallax` package.

kindoflew:
### Why
Draft of adding tests to eventually close #1474.

### What
This adds several things:
* A new demo app with `vite`, similar to the sandboxes
* `cypress` and `@testing-library/cypress`
* the actual tests -- `parallax.spec.js`

And also `package.json` scripts for `dev` and `test`. Everything new is inside `packages/parallax`.

### Checklist

- [ ] Documentation updated: n/a
- [ ] Demo added: sort of, but n/a
- [ ] Ready to be merged

#### CI
`cypress` requires a dev server to run tests. This can be manually done with two terminals open -- one for dev and one for `cypress`. But there's also [start-server-and-test](https://github.com/bahmutov/start-server-and-test) which would allow us to add it to the `release` pipeline (or at the very least, make things easier for contributors). I don't know how important that is, but it might be a nice-to-have? (Related -- should I add a script to `package.json` in the root folder to be able to run these tests from there as well?)

#### Actual tests
Currently, there are two main tests -- one each for a vertical and horizontal parallax app. In trying to make them actually E2E they just scroll/click and then check that everything is where it should be (by comparing their `transform: translate3d` values to what I know they should be, etc). But, after discovering the bug that lead to #1530 it seems that might not suffice (i.e. I think these tests might have passed even without that bugfix). Because of that I think it might be a good idea to add snapshot tests as well (but I still think testing the `translate` values is beneficial). I just wanted to get some general input before adding/removing more stuff.

Let me know if I'm totally off-base with anything, this is my first time attempting E2E tests (or really anything besides unit tests :sweat_smile:).


kindoflew:
I was researching which screenshot plugin to use and I came across cypress-io/cypress#9443. This could make for flaky tests because `Parallax` has resize listeners that will cause animated translation of the layers' position and size. 

For now, we can just stick with the direction we have and keep an eye on that issue and implement screenshots when it's fixed.

joshuaellis:
Hey, Sorry I missed the above comments in the main entry from your edits. Responses below:

>should I add a script to package.json in the root folder to be able to run these tests from there as well

I think that would be good, we could edit `yarn test` to be `yarn test:unit` or something like that, and yarn test should run _everything_.

I think your test approach is a good idea, the translate value is the numerical source of truth here, but appreciate it might not always be accurate the dom being fickle. So adding snapshots sounds like a great idea. We should also make sure we add variation to the horizontal and vertical, to test things like `sticky` etc. But sounds good.

Good catch on the issue in `cypress` I think we should see how it goes and play from there.

kindoflew:
I've added snapshots and a script to the root `package.json`. Each test (`vertical` and `horizontal`) has a normally translating layer (parent default `horizontal`), an opposite translating layer (if parent has `horizontal={true}`, then layer has `horizontal={false}`), a sticky layer, and a button for scrolling to the second page with `scrollTo`. 

Each `describe` function has two `it` functions -- one for `'should translate layers as expected'` which compares DOM snapshots and asserts `translate` values after scrolling. And one for `'should scroll to correct page with scrollTo'`. The Cypress docs suggested that it's best practice to have large `it` functions, as opposed to a lot of little ones, so that's the structure I went with. 

Are there any other variations you can think of that would be nice to test for?

package.json
```diff
@@ -41,9 +41,10 @@
     "watch": "preconstruct watch",
     "dev": "preconstruct dev",
     "demo:dev": "yarn workspace demo dev",
-    "test": "jest",
+    "test:unit": "jest",
     "test:cov": "jest --coverage",
     "test:ts": "tsc --noEmit",
+    "test": "yarn test:unit && cd ./packages/parallax && yarn test",
```
joshuaellis, line 47: can we remove the need for `cd` bringing `cypress` to the top level?

Also let's specifically name this `test:e2e` and then test can be `yarn test:unit && yarn test:e2e`

packages/parallax/cypress/support/commands.js
```diff
@@ -0,0 +1 @@
+import '@testing-library/cypress/add-commands'
```
joshuaellis, line 1: so could this go to the top level? `root/cypress`

packages/parallax/package.json
```diff
@@ -26,11 +26,23 @@
...
+  },
   "dependencies": {
     "@react-spring/shared": "~9.2.0",
     "@react-spring/web": "~9.2.0"
   },
   "peerDependencies": {
     "react": "^16.8.0  || ^17.0.0"
+  },
+  "devDependencies": {
+    "@testing-library/cypress": "^7.0.6",
```
joshuaellis, line 43: these dev deps should be in the root of the project, only actual deps are here in a monorepo 👍🏼

packages/parallax/demo/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
joshuaellis, line 1: maybe these files could be in `packages/parallax/test/*` to match the signature of the rest of the repo. Probably a good idea to add a little readme explaining this is for e2e tests too, what do you think?

joshuaellis (changes requested):
Tests look great! Just some thoughts around file structure that i've had, what do you think?

packages/parallax/package.json
kindoflew: Got it! Moved them and everything still works.

package.json
kindoflew: I've renamed the test scripts and also got rid of `cd` by using `yarn workspace @react-spring/parallax test`. Was `cd` the issue or is it that `/cypress` should be in the root?

packages/parallax/cypress/support/commands.js
kindoflew: Pretty sure that should work. Everything is orchestrated by `start-server-and-test` anyways so as long as the commands fed to that are right it should be fine. 

Or do you mean move the file like `/cypress/commands.js`? (This is also totally possible)

packages/parallax/demo/index.html
kindoflew: Yeah, both of those make sense to me. It isn't really a demo in that we don't want anyone messing with it unless they are writing new tests. And a `README` will double down on making that clear.

joshuaellis: exactly. ⭐

package.json
joshuaellis: I was just thinking in general that the cypress set up should be at the root of the mono repo 👍🏼

packages/parallax/cypress/support/commands.js
joshuaellis: see comment above, it'll probably clear it up a bit

package.json
kindoflew: Got it! Moved `cypress` folder and `cypress.json` to root.

packages/parallax/demo/index.html
kindoflew: Changed name and added `README` explaining it's for testing and how to generate new snapshots should someone contribute tests that alter the DOM.

joshuaellis (approved):
Genuinely 👏🏼  amazing 👏🏼  work 👏🏼  thank you so much.

kindoflew:
thanks! and no problem!

Issue #1430, opened 2021-04-17
@react-spring/three doesn't update all position parameters when passed to r3f mesh


midanosi:
## 🐛 Bug Report

I have a spring that I'm trying to use to make position parameter for a Box in r3f to go from e.g. [1,1,1] -> [2,2,2]
```js
const position = expandSpring.to((expandFactor) => [
  xInit + Math.sign(xInit) * expandFactor,
  yInit + Math.sign(yInit) * expandFactor,
  zInit + Math.sign(zInit) * expandFactor
])
```
But despite interpolating the spring to an array of values, when I pass it into <a.mesh position={interpolatedPosition} />, it only updates the x coordinate. If I change the interpolation fn to make x value static, e.g.
 ```js
[xInit, yInit + Math.sign(yInit) * expandFactor, zInit + Math.sign(zInit) * expandFactor]
```
then it'll now update only the y value.

I wasn't sure whether to raise this here or in the r3f repo! Sorry if I got the wrong place. Also aware that this might not be a bug, in which case may I ask for guidance on how to achieve what I'm aiming for?

## To Reproduce

See codesandbox

## Expected behavior

For all x, y, z parameters to be updated

## Link to repro (highly encouraged)

https://codesandbox.io/s/exploding-square-forked-for-react-spring-github-issue-dont-touch-n48b9

## Environment

- `@react-spring/three` v9.1.1
- `react` v17.0.2

Edit: changed sandbox link to stay static

midanosi:
... only 10 minutes later I came across this in r3f api readme: https://github.com/pmndrs/react-three-fiber/blob/master/markdown/api.md#piercing-into-nested-properties

So I'm now setting individual positions using `position-y` and that's doing the trick!

May I still ask the newbie question of why this is the case? If it's easy to answer that question?
This definitely felt like a gotcha moment that'd be easy for others to stumble across. I'd be happy to add a small README section in one of the repos if it'd be appropriate to do so.

joshuaellis:
Can you reduce your sandbox down just to highlight the core issue you're facing? There's too much noise for me to debug.

Side note, yes using `position-x` etc. will work, but I think `position` should be able to work too 🤔

midanosi:
Good point, sorry about that. Have gotten rid of all the clutter, same link here: https://codesandbox.io/s/exploding-square-forked-for-react-spring-github-issue-dont-touch-n48b9?file=/src/App.js

mysterybear:
[...]

PR #1550, opened 2021-06-06
fix: update all values of animated array (#1430)

midanosi:
### Why

Fixing bug https://github.com/pmndrs/react-spring/issues/1430

### What

Implemented suggested fix by a user in that thread, and wrote a unit test too.
I confirmed that the unit test failed in the expected way when the fix was not applied - i.e. spring of [10,20] is then updated to [11,21], but the resulting prop on animated div is still [11,20], because only the first animated value in the array was being updated.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution especially the test! I've merged this into another PR focussed on fixing some `r3f` related issues, so it will be released in one patch.

PR #1560, opened 2021-06-09
fix(support): incorrect type imports for rafz

sudkumar:
### Why

Fixes #1559 

### What

Import type declarations from `@react-spring/rafz` instead of `packages/rafz`

### Checklist

- Documentation updates: Not required
- Demo: Not needed
- Ready to be merged: Yes

PR #1561, opened 2021-06-09
chore: update inter-library depdency versions

zposten:
### Why

#1560 fixed a small typescript issue, but that fix unfortunately appears to not be used when installing `@react-spring/web@9.2.3`.

I believe the reason for that is the dependency versions of the various `@react-spring` packages were not updated.

This is from my `yarn.lock` file:

```
"@react-spring/web@^9.2.3":
  version "9.2.3"
  resolved "https://registry.yarnpkg.com/@react-spring/web/-/web-9.2.3.tgz#b9037719b489863ca2ad39caaacff6d315bbaea8"
  integrity sha512-dWRcgVDbO2UI9I03n/HVmCx9tY++Na+RwRzkzXv3E53BcFsjvnWGArnpj+xE/XgXiaII3ep2RmUj5jyYoukqGg==
  dependencies:
    "@react-spring/animated" "~9.2.0"
    "@react-spring/core" "~9.2.0"
    "@react-spring/shared" "~9.2.0"
    "@react-spring/types" "~9.2.0"
```

### What

I updated all of the `@react-spring` dependency versions from `9.2.0` to `9.2.3`. 

I'm not 100% certain what some of the packages like `@react-spring/zdog` or `@react-spring/konva` are, but it seems like all the version numbers are always kept consistent, and my `yarn install` after this change seemed to still work.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/packages/animated/package.json
+++ b/packages/animated/package.json
@@ -28,8 +28,8 @@
     "Josh Ellis (https://github.com/joshuaellis)"
   ],
   "dependencies": {
-    "@react-spring/shared": "~9.2.0",
-    "@react-spring/types": "~9.2.0"
+    "@react-spring/shared": "~9.2.3",
+    "@react-spring/types": "~9.2.3"
   },
   "peerDependencies": {
     "react": "^16.8.0  || ^17.0.0"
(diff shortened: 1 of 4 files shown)


joshuaellis:
I dont think this is the solution to your issue, this is a patch. AFAIK ~ means it'll only ever install version in this pattern 0.0.X, therefore npm should try to install `9.2.3` of the library even though it says `9.2.0` see [here](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept). I would try remove `@react-spring` packages and reinstall them. This will update the `yarn.lock` I believe.

sudkumar:
I agree with @joeporpeglia. This should do it. In my case, I updated the version in my package.json and ran npm install.

Issue #1582, opened 2021-06-23
TypeError: Cannot create property 'Symbol(AnimatedComponent)


Vazgen7788:
## 🐛 Bug Report

In case `Array.prototype` has a key which value type is `boolean` react spring throws an error. 

```
react-spring-animated.cjs.dev.js:367 Uncaught TypeError: Cannot create property 'Symbol(AnimatedComponent)' on boolean 'true'
```

## To Reproduce

- `Array.prototype['prop'] = true`
- Boot react application

## Expected behavior

React spring should not use all properties in `Array.prototype`

## Environment

- `react-spring` ^9.2.3
- `react` v17.0.2

Vazgen7788:
https://codesandbox.io/s/sad-driscoll-xtotq?file=/public/index.html

### Commenting line 27 will fix issue. 
<img width="546" alt="Screen Shot 2021-06-23 at 18 08 57" src="https://user-images.githubusercontent.com/8538941/123111589-159bc600-d44e-11eb-8a4c-5554fb8ccf9f.png">

joshuaellis:
I'm gonna be honest, this is a really weird issue that i've not seen before, we're not using `Array.prototype` in that file, we do however use `Array.from`, but if you want this "bug" fixing then I think you'd have to do some investigation yourself...

It would be helpful if you could explain some context as to why you want to use `Array.prototype`?

Vazgen7788:
Honestly, I don't want to use `Array.prototype`. 
We use react in Ember application. Ember mutates `Array.prototype`, and react app stops working. 
Will explore to find solution.

joshuaellis:
I wonder if it's the use of `Symbol`, i've never used it personally.

PR #1585, opened 2021-06-24
Fix issue 1582

Vazgen7788:
### What

Fix issue [1582](https://github.com/pmndrs/react-spring/issues/1582)

### Why

[eachProp](https://github.com/pmndrs/react-spring/blob/26bd8f153dd9284915b6fbc2bc299320c44b593d/packages/shared/src/helpers.ts#L47) function accepts argument [primitives](https://github.com/pmndrs/react-spring/blob/26bd8f153dd9284915b6fbc2bc299320c44b593d/targets/web/src/primitives.ts#L2) which type is `Array`. 
Iteration works over `Array.prototype` and in case there is a `boolean` it throws an error. It should use only values from [primitives](https://github.com/pmndrs/react-spring/blob/26bd8f153dd9284915b6fbc2bc299320c44b593d/targets/web/src/primitives.ts#L2) and should not iterate over `Array.prototype`. 

Screen Shot 2021-06-24 at 15 17 23

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/packages/shared/src/helpers.ts
+++ b/packages/shared/src/helpers.ts
@@ -54,7 +54,9 @@ export function eachProp<T extends object, This>(
   ctx?: This
 ) {
   for (const key in obj) {
-    fn.call(ctx as any, obj[key] as any, key)
+    if (obj.hasOwnProperty(key)) {
+      fn.call(ctx as any, obj[key] as any, key)
+    }
   }
 }
 


joshuaellis (approved):
thanks for the contribution! ⭐

Issue #1532, opened 2021-06-02
useSprings does not properly clear controllers when length changes in React.StrictMode


tarngerine:
## 🐛 Bug Report
When changing the length provided to useSprings in combination with a propsFn, the internal SpringRef does not properly dispose of the old refs when the array changes.

## To Reproduce

See CSB

## Expected behavior

See CSB, while the springs are correctly oscillating between lengths of 8 and 9, the api.current controllers always increase.
<img width="515" alt="image" src="https://user-images.githubusercontent.com/1629901/120428435-4306bf80-c339-11eb-97b0-935112f1c67d.png">

## Link to repro (highly encouraged)

https://codesandbox.io/s/elastic-hofstadter-m48zc?file=/src/UserTiles.jsx

## Environment

- `react-spring` v9.2.1
- `react` v17.0.2

joshuaellis:
This looks like it was brought about in this commit – https://github.com/pmndrs/react-spring/commit/9176c5a87a4eb3f1e184dec0b68b789aca50f542 Might have been when I used `useRef` to assign the `Controllers`.

This fix _shouldn't_ be too hard, if youre interested in giving it a go and contributing! ⭐

tarngerine:
might have some time this week, thanks!

tarngerine:
hm, i added some tests to useSprings.test.tsx but it seems to be passing locally. any thoughts on this?

<img width="856" alt="image" src="https://user-images.githubusercontent.com/1629901/120511488-15009a00-c390-11eb-8917-eb49a9769028.png">

tarngerine:
oh hm, if im reading it correctly, these tests create a new useSpring hook instance every time `update` is called?

joshuaellis:
No it doesn't because you don't destroy the spring when you update the props because of internal state etc.

tarngerine:
ah, i see...

so without making any changes to the useSprings hook, i downloaded the CSB as a zip and added it to the demo page and it works fine
<img width="2032" alt="image" src="https://user-images.githubusercontent.com/1629901/120516472-036dc100-c395-11eb-9c8d-e3ffc3fca962.png">

i did have to change the package import to `@react-spring/web` 

any clues on how this difference in behavior on prod 9.2.1 vs local monorepo might happen?

joshuaellis:
It's `StrictMode`. Add it to `ReactDOM.render` and the bug is there.

joshuaellis:
[...]

PR #1597, opened 2021-06-29
[RFC] fix: useSprings controller clear refs properly when length changes in React.StrictMode #1532

tarngerine:
### Why
fixes https://github.com/pmndrs/react-spring/issues/1532#issuecomment-870784013

### What
RFC

### Checklist

- [ ] Documentation updated (N/A)
- [ ] Demo added
- [ ] Ready to be merged


demo/src/index.tsx
```diff
@@ -3,4 +3,11 @@ import ReactDOM from 'react-dom'
...
 
-ReactDOM.render(<App />, document.getElementById('root'))
+ReactDOM.render(
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+  document.getElementById('root')
+)
+
+// ReactDOM.render(<App />, document.getElementById('root'))
```
joshuaellis, line 13: Lets get rid of this comment. It's probably better to make sure the library works in Strict Mode.

demo/src/App.jsx
```diff
@@ -52,6 +53,7 @@ const links = {
   'parallax-vert': ParallaxVert,
   'parallax-sticky': ParallaxSticky,
   three: SimpleThree,
+  test: Test,
```
joshuaellis, line 56: Can we name this something more intuitive, I guess this is all about adding and subtracting springs? So something along that line.

joshuaellis (changes requested):
Looking at the changes, the tests pass, I think if we just do those little changes, we can release this as a patch. Would you be happy to offer some support if this leads to unexpected side effects in people's code? although I'm hoping it wont 🙏🏼

demo/src/App.jsx
tarngerine: Do you think we should include this demo or just rely on the new tests? It's not really at the calibre of the rest of the demos, and I just grabbed it from a discord user who had posted this bug originally.

tarngerine: From ericvu in #react-spring

tarngerine: i think this demo app is super confusing since it's half-broken, do you think its okay to remove this demo app and just rely on the tests?

tarngerine:
ok, i removed the demo app bc i dont think its necessary as we have implemented the tests, and removed the commented non-strict-mode line

happy to provide any support post launch, altho i will be "out of office" the first week of august

lmk good next steps, thanks!

joshuaellis:
Hey @tarngerine sorry, I've been so snowed under with work at the moment. Yeah I think upon reflection, you're probably right, let's remove the app and keep the tests, thats the point of tests after all?

joshuaellis (approved):

PR #1600, opened 2021-06-29
fix: add `immediate` to payload in useTransition hook

### LINKED ISSUE #1581: The `immediate` property no longer works with `useTransition` hook
## 🐛 Bug Report

The `immediate` property no longer works with `useTransition` hook.

## To Reproduce

See CodeSandbox. Use the `useTransition` hook with `immediate: true` passed in as a property.

## Expected behavior

No animations, transitions happen immediately.

https://codesandbox.io/s/react-spring-usetransition-immediate-bug-4vtob?file=/src/App.tsx

If you change the version to `9.0.0-rc.3` you will see that the transitions are immediate.

## Environment

- `react-spring` v9.x (non-rc)
- `react` v17.0.2

wyze:
### Why

This change does fix #1581. I'm not sure if this is the right change though, given the comment above the code.

### What

Add `immediate` back to the array of default props to allow `useTransition({ immediate: true })` to be respected again.

### Checklist

- [ ] Demo added
- [x] Ready to be merged

### Notes

If this is indeed a proper fix, I'd like to add a test somewhere. Any help with that would be great.

### DIFF
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -95,6 +95,7 @@ export const getDefaultProps = <T extends Lookup>(
  */
 export const DEFAULT_PROPS = [
   'config',
+  'immediate',
   'onProps',
   'onStart',
   'onChange',
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -95,7 +95,6 @@ export const getDefaultProps = <T extends Lookup>(
  */
 export const DEFAULT_PROPS = [
   'config',
-  'immediate',
   'onProps',
   'onStart',
   'onChange',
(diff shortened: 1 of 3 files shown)


joshuaellis:
I'm not sure this should be added to default props... if you add `immediate: true` to the phases it works as [intended](https://codesandbox.io/s/react-spring-usetransition-immediate-bug-forked-unn98?file=/src/App.tsx). If we want it to work when passing in the regular `transitionProps` then we probably need to include it in `payload`

https://github.com/pmndrs/react-spring/blob/055eb70733305b7bf2cbc555a02d54816ddeeecb/packages/core/src/hooks/useTransition.tsx#L240

wyze:
Thanks for the direction! I've updated the code and added a test.

Issue #1649, opened 2021-07-30
React-Spring's initialization breaks on some polyfilled Arrays (specifically Ember builds)


brownstein:
## 🐛 Bug Report

Ember polyfills Array for observability, but that polyfill's addition of a key to the standard prototype breaks React-Spring's primitive iteration in the `createHost` method. I'm going to write a PR to fix this using an `Array.isArray` check to switch the iteration method to a standard `for` loop instead of a `for ... in ...` iteration.

## To Reproduce

Import React-Spring in an Ember build, or any build that unsafely polyfills Array in a way that breaks standard for-in iteration.

## Expected behavior

Imports of React-Spring do not break in Ember builds.

## Link to repro (highly encouraged)

(Please provide either a [CodeSandbox demo](https://codesandbox.io) or an example GitHub repo.)

## Environment

- `react-spring` v8.x.x
- `react` v16.8.x (or `react-native` v0.58.x)

PR #1650, opened 2021-07-30
Fix iteration across arrays to avoid conflict with polyfilled Array variants

brownstein:
### Why
Resolves https://github.com/pmndrs/react-spring/issues/1649: using for-in to traverse Arrays is normally safe, but with Ember's observability polyfills, it causes `createHost` to break on primitive iteration.

### What

This fix swaps iteration to use a for loop when the traversed object is an array.

### Checklist

- [X] Ready to be merged

### DIFF
--- a/packages/shared/src/helpers.ts
+++ b/packages/shared/src/helpers.ts
@@ -53,6 +53,12 @@ export function eachProp<T extends object, This>(
   ) => void,
   ctx?: This
 ) {
+  if (is.arr(obj)) {
+    for (let i = 0; i < obj.length; i++) {
+      fn.call(ctx as any, obj[i] as any, `${i}`)
+    }
+    return;
+  }
   for (const key in obj) {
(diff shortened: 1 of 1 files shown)


michaeldrob (approved):

joshuaellis:
Shouldn't we remove the old code for this? I'll look at this over the coming days. Thank you for the contribution!

Do you have a link about this Ember polyfill injection?

brownstein:
@joshuaellis I don't think we should remove the old code; the old iterator is designed to work with Objects, whereas this path is just for Arrays. Here are some docs on Ember's injection behavior: https://guides.emberjs.com/v1.11.0/configuring-ember/.

joshuaellis (approved):
Yeah you're right, sorry, I saw `obj` and not `arr` and got confused. Thanks for the fix! ⭐

Issue #1639, opened 2021-07-21
animated does not work as expected on components also wrapped with forwardRef


gpoole:
## 🐛 Bug Report

Using `forwardRef` on a component seems to stop `animated` from updating the component in response to the props received from `useSpring`.

## To Reproduce

 The following usage causes the issue described:

```js
const AnimatedNotWorking = animated(
  forwardRef(({ value }, ref) => {
    return <div ref={ref}>{value}</div>;
  })
);
```

## Expected behavior

The component wrapped in `animated` should be re-rendering in reaction to changing animated props, which it does if `forwardRef` isn't used, for example the following works as expected:

```js
const AnimatedWorking = animated(
  ({ value }) => {
    return <div>{value}</div>;
  }
);
```

## Link to repro (highly encouraged)

https://codesandbox.io/s/charming-mestorf-hptid?file=/src/App.js

In the sample, `AnimatedWorking` is working as expected and re-rendering continuously as the animated value changes, `AnimatedNotWorking` is only rendering once.

## Environment

- `react-spring` v9.2.4
- `react` v17.0.2

kindoflew:
After a little digging I found [this article](https://dev.to/justincy/using-react-forwardref-and-an-hoc-on-the-same-component-455m) explaining how `forwardRef` and HOCs can behave a little weird when used together (`animated` being an HOC). 

I hobbled together a [version of your example](https://codesandbox.io/s/modern-pine-iyl4c?file=/src/Test.js) using the solution in that article and it seems to work (the `background-color` is set to `'pink'` using the `ref` in `App.js` and the value is still animated correctly). 

So, I *think* this is intended (if unexpected) behavior for React.

gpoole:
Thanks for looking into it @kindoflew. The article is describing workaround for HOCs that don't forward refs so the ref is being dropped, but as the article mentions:

> We could update our HOC to pass on refs, but we might want it to be used with components that don't accept refs. So what can we do?

And in this case `animated` is actually correctly forwarding the ref to the wrapped component, so refs are working as expected with `animated` components. The issue I'm reporting is different though. In my repro the working component constantly renders with the changing spring value, whereas the broken one renders once on mount and stops, which is the problem.

Just to add a bit of detail: the broken component _does_ render in response to other prop updates, it just doesn't seem to be updated at all by the animated spring properties changing. I've also noticed in my app that if the component happens to be re-rendered by other props changing it render with a snapshot of where the spring properties were at at the time.

[...]

PR #1671, opened 2021-08-31
fix: handle ParallaxLayers inside fragments (#1667)

### LINKED ISSUE #1667: Parallax sticky elements don't work when wrapped in div or fragment
## 🐛 Bug Report
If a ParallaxLayer with the sticky property passed is wrapped in a div or fragment, the element disappears for the time it should be "sticking" on screen. Removing any enclosing div/fragments returns to correct behavior. This is a problem on a larger site I'm developing where I would like to split some of the layers up into subcomponents, which necessitates being able to include extra divs/fragments.  

## To Reproduce
Put a fragment around a ParallaxLayer w sticky property set. It can enclose more than just that one ParallaxLayer but the problem shows up w just the one enclosed. The sticky element is visible before and after its designated sticky phase, but disappears during. 

## Expected behavior
A sticky element should remain visible during the time it is sticky

https://codesandbox.io/s/determined-merkle-yi6ee?file=/src/App.tsx

This is a fork of the [official example](https://codesandbox.io/s/github/pmndrs/react-spring/tree/master/demo/src/sandboxes/parallax), with the addition of a <></> around the "I'm a sticky layer element". 

## Environment
My environment where I first encountered the issue 
- "@react-spring/parallax": "^9.2.4",
- "react": "^17.0.2",
- "next": "11.01.1",

kindoflew:
### Why

Resolves #1667 

### What

I've added a check where `children` is mapped to see if the `child` is a `React.Fragment`. If it is, we also map over the `fragment`'s `children`. 

### Checklist

- [ ] Ready to be merged

A couple related questions/thoughts:

* I threw together a very simple `isReactFragment` function just to confirm this works. Should we use the `react-is` package instead? One benefit of that is that it might also aid in figuring out a solution to #1639 (if we decide that we should do something about that). Also, it will be less brittle because we won't be relying on React internals.
* I'm not sure if the way in which I map over the `children` of a `fragment` is ideal.

Hopefully, one day, `overflow: clip` will solve most of the workarounds we use for `sticky`.

### DIFF
--- a/packages/parallax/src/index.tsx
+++ b/packages/parallax/src/index.tsx
@@ -14,6 +14,13 @@ function getScrollType(horizontal: boolean) {
   return horizontal ? 'scrollLeft' : 'scrollTop'
 }
 
+function isReactFragment(node: any) {
+  if (node.type) {
+    return node.type === React.Fragment
+  }
+  return node === React.Fragment
+}
+
 const START_TRANSLATE_3D = 'translate3d(0px,0px,0px)'
 const START_TRANSLATE = 'translate(0px,0px)'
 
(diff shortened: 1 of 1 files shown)


kindoflew:
I forgot to mention, this fix also assumes that we should only allow`ParallaxLayers` or `fragments` to be direct `children` of `Parallax` (so no `div`s, etc.) and we should probably document that.

kindoflew:
Continuing from [this comment](https://github.com/pmndrs/react-spring/issues/1667#issuecomment-917201877), I'm not sure that this idea is a complete fix for this bug. I also think it'd be too much to check everything a user places inside `Parallax` to make sure that it doesn't have `ParallaxLayer`s as `children`.

Not really sure what a good solution would be. Especially considering that there are two CSS rules (`overflow: clip` and `contain: paint`) that would eliminate issues like this, except that neither is supported by Safari (or IE, but, you know).

But since doing nothing is (probably?) worse than a half-finished fix, maybe I should just finish this PR and we can use it as a stop-gap, with the above mentioned caveat (only `ParallaxLayer`s and `fragment`s as direct `children`) and just continue waiting.

joshuaellis:
I think for now, just using the function you've created is a better solution. If you can finish the PR, and let me know why you don't think its the best way to do it, I can have a look and see what I can come up with based on those issues 👍🏼

kindoflew:
Also, not sure if the Vercel deployment failing was something I did, but e2e tests were passing locally before I pushed.

joshuaellis:
No, you didn't break it. I don't think I set it up correctly before, but now I have.

If you could update the docs at the same time that would be great! ⭐

joshuaellis (dismissed):

Issue #1572, opened 2021-06-15
Children prop missing from React Native Components


BraveLilToaster:
## 🐛 Bug Report

I am experiencing a Typescript issue when trying to create animated `View` or `Text` in React Native. The returned components from the animated function are missing a `children` prop.

## To Reproduce

1) Create a component by calling `animated()` with either the `View` or `Text` components from `react-native`.

```Javascript
import { a } from "@react-spring/native";
import { Pressable, Text, View } from "react-native";

const AnimatedView = a(View);
```

2) Wrap any JSX with the animated component.

```JSX
<AnimatedView>
  <Text>this is broken</Text>
</AnimatedView>
```
3) The Typescript compiler should throw an error about a missing property `children`.

>  Type '{ children: Element; accessibilityRole: "link"; style: { width: SpringValue<number>; }; }' is not assignable to type 'IntrinsicAttributes & AnimatedProps<ViewProps & RefAttributes<View>>'. Property 'children' does not exist on type 'IntrinsicAttributes & AnimatedProps<ViewProps & RefAttributes<View>>'.

## Expected behavior

I would expect components like `View` and `Text` to have the property `children` as an optional prop.

## Link to repro (highly encouraged)

https://codesandbox.io/s/blissful-monad-fips9?file=/src/App.tsx

## Environment

- `react-spring` v9.2.3
- `react` v17.0.2
- `react-native` v0.64.2
- `react-native-web` v0.16.5

joshuaellis:
You're welcome to submit a PR for this, I imagine it's a relatively straight forward fix! 😄

gorillatron:
Also experiencing this

eladchen:
@joshuaellis Any pointers you want to share? This is not as straightforward as it sounds.

PR #1705, opened 2021-10-04
fix(native): add children prop

CodyJasonBennett:
### Why

As reported in #1572, react-spring does not accept props derived directly from React, like props.children.

### What

This PR adds the children prop to AnimatedComponent's types.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/targets/native/src/animated.ts
+++ b/targets/native/src/animated.ts
@@ -1,4 +1,4 @@
-import { ForwardRefExoticComponent } from 'react'
+import { ForwardRefExoticComponent, ReactNode } from 'react'
 import { ViewStyle, RecursiveArray } from 'react-native'
 import {
   AssignableKeys,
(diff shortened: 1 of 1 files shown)


targets/native/src/animated.ts
```diff
@@ -19,9 +19,10 @@ export type WithAnimated = {
 } & AnimatedPrimitives
 
 /** The type of an `animated()` component */
-export type AnimatedComponent<
-  T extends ElementType
-> = ForwardRefExoticComponent<AnimatedProps<ComponentPropsWithRef<T>>>
+export type AnimatedComponent<T extends ElementType> =
+  ForwardRefExoticComponent<
+    AnimatedProps<ComponentPropsWithRef<T>> & { children: ReactNode }
```
LinusU, line 24: This change seems to have broken the following code for me:

```js
<animated.View style={{ opacity }} onTouchEnd={handleDismiss} />
```

It now gives this error:

```text
error TS2741: Property 'children' is missing in type '{ style: { opacity: SpringValue<number>; }; onTouchEnd: () => void; }' but required in type '{ children: ReactNode; }'.
```

I think that this should have been:

```suggestion
    AnimatedProps<ComponentPropsWithRef<T>> & { children?: ReactNode }
```

CodyJasonBennett: Indeed, could you open a PR?

LinusU: Yeah I should probably have linked that one, sorry 😅 

https://github.com/pmndrs/react-spring/pull/1925

It's merged now 🚀

PR #1719, opened 2021-10-19
Fix 'useSpring'  'SVG Filter' demo

hualu00:
### Why
Currently 'SVG Filter' demo doesn't work when you click, in which a toggle effect is expected.  
https://codesandbox.io/s/github/pmndrs/react-spring/tree/master/demo/src/sandboxes/svg-filter

### What
change the useSpring parameter from function to object.

### Checklist

- [x] Documentation updated
- [x] Demo added
- [x] Ready to be merged


joshuaellis:
refactor: use deps array in example

PR #1734, opened 2021-11-03
docs: Describe combining ref forwarding with `animated`

### LINKED ISSUE #1639: animated does not work as expected on components also wrapped with forwardRef

gpoole:
### Why

Resolves #1639 by documenting an issue with combining `animated` with `forwardRef` and explaining how to work around it.

### What

I've updated the documentation to describe the problem in detail with examples and describe a workaround.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/docs/src/pages/basics.mdx
+++ b/docs/src/pages/basics.mdx
@@ -380,3 +380,48 @@ useEffect(() => {
   api.start({ x: 100, y: 100 })
 }, [])
 ```
+
+### Forwarding refs and the `animated` HoC
+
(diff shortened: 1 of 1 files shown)


joshuaellis (approved):
People who update documentation are the real MVP. Thanks for the contribution! ⭐

gpoole:
Thanks! Thanks @kindoflew for your research and the workaround.

Issue #1766, opened 2021-11-23
Issue with rerender when using the Parallax enabled property


Lucasdsk:
## 🐛 Bug Report

I'm trying to enable and disable the Parallax scroll dynamically through the **`enabled`** property.

Doc.: https://react-spring.io/components/parallax#parallax-1

The issue is that I'm getting the error:

![image](https://user-images.githubusercontent.com/7729844/143083000-3dd2ed60-3e3b-497b-9f68-ea6eebaf49c1.png)

This is breaking the whole page behavior, the scroll stops working normally. 

## To Reproduce

Change the **`enabled`** property dynamically.

## Expected behavior

Enable and disable the scroll normally.

## Link to repro (highly encouraged)

I've created this example that's pretty similar to what I have in my project.
https://codesandbox.io/s/react-spring-parallax-7bt5t?file=/src/App.tsx

## Environment

- `react-spring` v8.0.27
- `react` v17.0.1
- `@react-spring/parallax`: "^9.3.0",

joshuaellis:
resolved by #1767

PR #1767, opened 2021-11-23
fix: update overflow css property without mixing shorthand and non-shorthand properties

Lucasdsk:
### Why

It fixes the following issue: https://github.com/pmndrs/react-spring/issues/1766

Also, you can check an example here:
https://codesandbox.io/s/react-spring-parallax-issue-1766-f4iow?file=/src/App.js

### What

I have changed the way the property `overflow` is provided to the Parallax component, instead of mixing the shorthand and the non-shorthand properties, I'm using only non-shorthand.

It will avoid this:
image

### Checklist

- [x] Demo added
- [ ] Ready to be merged

### DIFF
--- a/packages/parallax/src/index.tsx
+++ b/packages/parallax/src/index.tsx
@@ -321,7 +321,16 @@ export const Parallax = React.memo(
       return () => window.removeEventListener('resize', onResize, false)
     })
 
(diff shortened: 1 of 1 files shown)


joshuaellis (approved):
Thanks for the fix @Lucasdsk! Great work 💪🏼

PR #1771, opened 2021-11-30
Feature/add container refs

### LINKED ISSUE #1770: Can't get property container ref element - React-spring v9. ParallaxLayers. useRef
## Feature Request

Theres a small regression in what the refs contained when passed to Parllax components i.e. you don't get to access the actual nodes. It would be helpful to re-introduce these and add this to the documentation. They could be called `container` as they previously were, but i'm also open to suggestions of a more intuitive name.

### Discussed in https://github.com/pmndrs/react-spring/discussions/1769

<div type='discussions-op-text'>

<sup>Originally posted by **pvldsgn** November 24, 2021</sup>
Hello everyone!

I'm not sure if this belongs to the category of questions, because it looks like a bug:

When i want add event scroll to track its position on the page, i used useRef and almost everything worked, except for the smoothness of the scrollToTop event, but that's not about it now. Then I got the value of useRef as an object indicating what kind of object it was and could manipulate it (I needed to get the value of ref.current.container)

Was:
Снимок экрана 2021-11-24 в 15 53 28
console.log(ref)
console.log(ref.current.container)

At that time I used react-spring@8.0.27 with that dependency
`import { Parallax, ParallaxLayer } from 'react-spring/renderprops-addons'`

But when i needed add sticky container it's not worked
And I had to turn to the example from the react-spring documentation

There I encountered other dependencies
`import { Parallax, ParallaxLayer } from '@react-spring/parallax'`
With which the sticky container worked, but the opportunity to get the correct ref was broken

Become: 
object itself disappeared and became impersonal
Снимок экрана 2021-11-24 в 15 52 54
console.log(ref)
console.log(ref.current.container)

Question:
How do I get the correct ref in this case?

Thank's for any help!

</div>

kindoflew:
### Why

Resolves #1770.

### What

Added the DOM `ref`s for the outer and inner `div` containers -- `container` and `content` respectively (as they were named pre-`9.x.x`). Also, updated docs and `README` to reflect changes.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

For the docs, I updated the organization a little bit and made a "`ref` Properties" section that includes these two `ref`s, as well as the `scrollTo` method. It just seemed to stand out a little bit more, as opposed to a list (although, I suppose it could also be a table 🤔 ).

### DIFF
--- a/docs/src/pages/components/parallax.mdx
+++ b/docs/src/pages/components/parallax.mdx
@@ -52,16 +52,34 @@ import { Parallax, ParallaxLayer } from '@react-spring/parallax'
 | horizontal? | boolean       | Whether or not the container scrolls horizontally. Defaults to `false`.          |
 | innerStyle? | CSSProperties | CSS object to style the inner `Parallax` wrapper (not the scrollable container)  |
 
-### `scrollTo`
+### `ref` Properties
 
-`Parallax` also has a `scrollTo` function for click-to-scroll. It takes one paramater: the number of the page to scroll to
-(Pages are zero-indexed so `scrollTo(0)` will scroll to the first page, etc).
+`Parallax` also has a few useful properties that you can access using a `ref`:
+
+```jsx
+const ref = useRef()
+...
+<Parallax ref={ref}>
+```
+
+#### `ref.current.scrollTo`
+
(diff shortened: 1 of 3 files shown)


joshuaellis:
Just pointing this at release/9.4.0

I will release this on beta tonight, cc @pvldsgn

joshuaellis (approved):

pvldsgn:
I look forward, thank you @joshuaellis, @kindoflew

PR #1773, opened 2021-12-01
feat: add exitBeforeEnter prop to useTransition

### LINKED ISSUE #1064: Add "exitBeforeEnter" prop to useTransition
## 🚀 Feature Proposal
Add `exitBeforeEnter` property to `useTransition` hook. This property would be an optional boolean that when set will wait for "leave" animation to complete before starting "enter" animation of new element. Ideally we would also not need to worry about layout issues here and implement "hacks" like absolute positioning.

## Motivation
I came across an issue similar to one described here https://github.com/react-spring/react-spring/issues/583
Essentially I want to create a page transition, where "initial" page fades out and only after it is not visible anymore "secondary" page fades in. This should be achievable without complex logic and delays. 

Idea is taken from framer motion https://www.framer.com/api/motion/animate-presence/#animatepresenceprops.exitbeforeenter

I am down to try and implement this, but I am not sure where to start and if this is something that falls into the scope of the library.

## Example
```ts
const transition = useTransition(toggle, {
   from: { opacity: 0 },
   enter: { opacity: 1 },
   leave: { opacity: 0 },
   exitBeforeEnter: true
});

// ...

transition((style, props) => {
  return <a.div style={style}>{toggle ? 🟢 : 🔴}</a.div> 
})
```

joshuaellis:
### Why

This PR adds the `exitBeforeEnter` prop to the `useTransition` hook as requested here #1064. It also adds a new demo showcasing the new prop used with a `SpringRef`.

### What

Adds the prop and splits the changes into exiting and not is `exitBeforeEnter = true`. This works very well when no `SpringRef` is passed to the `useTransition` hook. 

The problem when using a ref is the ref parent component can't be flagged for re-render so the users API cannot call start to bring in the entering animation after the exiting has left. To solve this issue, we flag a `forceChange` when `forceUpdate` is called so instead of updating the queue of the passed ref, we instead run the animation straight away. 

This is inline with how calling `start` works with the prop turned off, the animations overlap – leave. & enter happen at the same time. So I think this is an acceptable solution.

Also added a test for this feature.

### Checklist

- [ ] Documentation updated
- [x] Demo added
- [x] Ready to be merged


joshuaellis:
I'll write up the documentation when the PR is approved, just incase something needs editing for whatever reason.

packages/core/src/hooks/useTransition.tsx
```diff
@@ -300,13 +306,35 @@ export function useTransition(
         }
         // Force update once idle and expired items exist.
         if (idle && transitions.some(t => t.expired)) {
+          /**
+           * Remove the exited transition from the list
+           */
+          exitingTransitions.current.delete(t)
```
kindoflew, line 312: should this be in the `if (exitBeforeEnter)` check below? It doesn't seem to be affecting the functionality in the demo or anything, but as far as I can tell, `exitingTransitions` doesn't get populated unless `exitBeforeEnter === true` (I also may be missing something though 😅 ).

kindoflew (commented):
Looks good to me! New demo looks great too.

I only had the one question in my comment and also there are a few commented out `console.log`s that I don't know if you want to keep in there.

packages/core/src/hooks/useTransition.tsx
joshuaellis: Good point. But if someone changes the prop we might still have it I think... so maybe we should be checking if t exists. Although according to MDN it just returns `false` if it didn't exist.... 

So it won't break, but maybe it's a more explicit thing to do?

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete

kindoflew: Thats a good point about changing the prop -- since later in the hook `exitingTransition.scurrent.size` is used to check in the `useLayoutEffect`. I assume we'd want it to be empty no matter what then? And if it just returns `false` it won't affect anything else.

I think with that in mind it *is* kind of explicit -- "We want this removed when done, no matter what."

joshuaellis:
Thanks for catching those rogue logs! I've removed them so going to move forward with this PR.

PR #1780, opened 2021-12-08
update readme to use the right rafz

arzafran:
### Why
Minor documentation fix to avoid confusion and using other rafz already installed.

### What
added the correct path to the rafz library under @react-spring

### Checklist

- [x] Documentation updated
- [x] Ready to be merged


joshuaellis (approved):

PR #1781, opened 2021-12-09
Update use-transition.mdx

david-buck:
Update "Transitioning between routes" example to use latest React Router syntax

### Why

React Router has changed a little since v5. This is to keep the docs up to date for people who are using the latest version of React Router.

### What

Updated the use-transition example for animating routes to use React Router 6's updated syntax https://reactrouter.com/docs/en/v6/api#routes-and-route

### Checklist

- [ x ] Documentation updated
- [ x ] Ready to be merged


joshuaellis (approved):
Awesome! thank you for the contribution ⭐

PR #1783, opened 2021-12-09
`skipAnimation` not working as expected with `delay` prop

martyan:
### Why

As seen in [this doc](https://react-spring.io/guides/testing#skipping-animations), `skipAnimation` prop should: `skip animations and jump straight to the 'to' value`. 
Which is what it does. But it applies the delay first (when defined ofc). 

Test below is based on the one from the linked docs. Only added `delay` to the animation. I want to test the `to` value as in the original test.
```
Globals.assign({ skipAnimation: true })
...
it('sets opacity to 1', () => {
  const { container } = render(<DelayedThing delay={2000} />)
  expect(container).toHaveStyle({ opacity: 1 })
});
```
Test fails because the delay is applied.

If I expect `skipAnimation` to finish all animations immediately, I would also expect `delay` not being applied.

Related discussion & repro code:
https://github.com/pmndrs/react-spring/discussions/1779

More docs:
https://aleclarson.github.io/react-spring/v9/#The-skipAnimation-global

### What

Added `skipAnimation` check to `scheduleProps`.


joshuaellis:
> Test below is based on the one from the linked docs. Only added delay to the animation. I want to test the to value as in the original test.

Sorry, is this a test that already exists or one you made that thus failed because of the bug? If latter, it would be great to include this text so we're more confident in the future!

martyan:
[Here](https://codesandbox.io/s/react-spring-skipanimation-test-fail-bdeid?file=/src/__tests__/EaseIn.js) is CSB with both tests.

The first one is example from [react-spring docs](https://react-spring.io/guides/testing#skipping-animations).
The other one tests the same thing (`opacity` should be `1`) but with delay. The test should pass as the `skipAnimation` is set. But it fails due to the `delay` being applied and thus `opacity` is still `0`.

joshuaellis (approved):

joshuaellis:
Sorry, I more meant about including a test or two in our suite, which I have now done.

martyan:
Awesome, thank you! :fire:

Issue #1732, opened 2021-11-02
AnimationFrame load CPU after component unmounted


Luchanso:
## 🐛 Bug Report

AnimationFrame continue calling after component destroyed. `stop()` not helped.

![image](https://user-images.githubusercontent.com/2098777/139920706-8a01827b-6e7a-4691-891d-4c9ad115fb6a.png)

![image](https://user-images.githubusercontent.com/2098777/139921498-78f26979-5c09-4c8f-a378-8aec1499d516.png)

## To Reproduce

https://codesandbox.io/s/exciting-shape-rqdk4?file=/src/App.js

## Expected behavior

AnimationFrame stop calling after component 
![image](https://user-images.githubusercontent.com/2098777/139921185-f39f617b-22c9-4582-acd0-5069784043a4.png)

## Link to repro (highly encouraged)

https://codesandbox.io/s/exciting-shape-rqdk4?file=/src/App.js

## Environment

- `react-spring` v9.3.0
- `react` v17.02

artman:
We're experiencing this same issue in production. Also happens on v9.2.4 where we upgrade from. After any single one react-spring animation has completed, animation frames are still being requested and CPU is at 2.5% even though no re-rendering is happening.

joshuaellis:
I haven't really looked into this yet, but off the top of my head, RAF is probably still being called after unmount, which would indicate its in the global scope or something similar

artman:
Just tested and this seems to be an issue with all 9.x.x versions.

joshuaellis:
> Just tested and this seems to be an issue with all 9.x.x versions.

This would make sense since v9 was a major re-write of most things incl. introducing `@react-spring/rafz`

srubin:
Would rafz be interested in a PR that stops the RAF loop if there are no scheduled timeouts or updates?

joshuaellis:
I'd be very interested in seeing it 🥳 sounds great!

joshuaellis:
this has been released on `v9.4.0-beta.3`

PR #1786, opened 2021-12-14
Stop RAF update loop when there are no pending tasks

srubin:
### Why

A quick test in Chrome shows that a no-op requestAnimationFrame update loop idles between 5-10% CPU usage ([codesandbox](https://codesandbox.io/s/cool-rui-15czz?file=/src/index.js:0-23)).

image

@react-spring/rafz and react-spring are susceptible to this behavior: once a task is scheduled with rafz, the update loop never stops even if there is never any other work scheduled. See this bug, which shows an example of this issue: https://github.com/pmndrs/react-spring/issues/1732

### What

This PR stops the raf update loop if there are no pending tasks. It does so using a task counting variable that was already present in the code but only used for testing. I added unit tests to verify that the update loop is running when necessary and stopped when able.

I also ran the code sample from https://github.com/pmndrs/react-spring/issues/1732 (and on another codebase) with this code change and verified that the raf update loop is not running indefinitely:

image

(left, before hitting "Destroy AnimationComponent"; right: after hitting "Destroy AnimationComponent")

I could use some guidance as to how to verify that this change is not negatively impacting or regressing anything else in the library.

### Checklist

- [ ] Documentation updated _(because this is just an implementation detail of rafz, is documentation required?)_
- [ ] Demo added _(in the form of unit tests, I guess?)_
- [ ] Ready to be merged

### DIFF
--- a/packages/rafz/src/index.ts
+++ b/packages/rafz/src/index.ts
@@ -35,12 +35,12 @@ raf.setTimeout = (handler, ms) => {
   let cancel = () => {
     let i = timeouts.findIndex(t => t.cancel == cancel)
     if (~i) timeouts.splice(i, 1)
-    __raf.count -= ~i ? 1 : 0
+    pendingCount -= ~i ? 1 : 0
   }
 
   let timeout: Timeout = { time, handler, cancel }
   timeouts.splice(findTimeout(time), 0, timeout)
-  __raf.count += 1
+  pendingCount += 1
 
   start()
   return timeout
@@ -107,6 +107,9 @@ raf.advance = () => {
 /** The most recent timestamp. */
 let ts = -1
 
+/** The number of pending tasks  */
+let pendingCount = 0
+
 /** When true, scheduling is disabled. */
 let sync = false
 
@@ -148,7 +151,7 @@ function update() {
   let count = findTimeout(ts)
   if (count) {
     eachSafely(timeouts.splice(0, count), t => t.handler())
-    __raf.count -= count
+    pendingCount -= count
   }
 
   onStartQueue.flush()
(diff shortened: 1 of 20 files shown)


packages/rafz/src/index.ts
```diff
@@ -201,7 +212,13 @@ function eachSafely<T>(values: Eachable<T>, each: (value: T) => void) {
 /** Tree-shakable state for testing purposes */
 export const __raf = {
   /** The number of pending tasks */
-  count: 0,
+  count(): number {
+    return pendingCount
+  },
+  /** Whether there's a raf update loop running */
+  isStarted(): boolean {
```
joshuaellis, line 219: I'm wondering if `isRunning` is a little clearer here?

packages/rafz/src/index.ts
```diff
@@ -201,7 +212,13 @@ function eachSafely<T>(values: Eachable<T>, each: (value: T) => void) {
 /** Tree-shakable state for testing purposes */
 export const __raf = {
   /** The number of pending tasks */
-  count: 0,
+  count(): number {
```
joshuaellis, line 215: It'd be helpful for me to understand why we've moved `count` to a pseudo `getter` function with a global variable.

joshuaellis (changes requested):
Just a few questions for my own understanding. But overall it looks great!

packages/rafz/src/index.ts
srubin: My understanding from the comments is that `__raf` is intended to be a container for data/functions that are used only in the tests. As of this PR, the `count` variable is now needed for actual control flow rather than just for test observability, so I moved it to a global.

packages/rafz/src/index.ts
srubin: Fixed

joshuaellis (approved):
Will release this in the upcoming 9.4.0 (it'll be released on beta later today). Thank you again!

srubin:
No problem, and thank you for the quick merge!

PR #1798, opened 2021-12-22
[Feat] css variables

### LINKED ISSUE #1651: CSS Variables With React Spring
## 🚀 Feature Proposal

react-spring should accept css variables as props and animate the same as if defined explicitly in the spring.

## Motivation

Theming requires styles and colors depending on theme type and using css vars is a common and easy way to map different colors & themes to one css variable. 

```
:root { 
 .default-theme {
   --box-shadow: 0px 0px 31px 1px rgba(98, 0, 238,1);
  }

  .dark-theme {
     --box-shadow: 0px 0px 10px 3px rgba(113,113,113,0.4);
  }

  .special-theme {
     --box-shadow: 0px 2px 10px 3px rgba(98,98, 238,1);
  }
}
```

## Example

I have a component that inherits box-shadow variables depending on the theme. The box-shadow should animate onHover.
If the box shadow is defined as a string, the animation works as expected. However, if I use my css variable - which works without animations - animations are broken. 

```
import React from 'react';
import PropTypes from 'prop-types';
import { useSpring, animated } from 'react-spring';
import './search.scss';

const Search = ({ change }) => {
  const [animation, api] = useSpring(() => ({
    from: { boxShadow: 'var(--box-shadow)' } }));

  return (
    <>
      <animated.input
        placeholder="🔎 Search..."
        className="search"
        onChange={change}
        style={animation}
        onMouseOver={() => api.start({ boxShadow: 'var(--box-shadow-hover)' })}
        onMouseLeave={() => api.start({  boxShadow: 'var(--box-shadow)' })}
        onFocus={() => console.log('focus')}
      />
    </>
  );
};

Search.propTypes = {
  change: PropTypes.func.isRequired,
};

export default Search;
```

### Problem 
Without being able to use css variables, the animating colors cannot be dynamically rendered depending on the theme.

joshuaellis:
### Why

* resolves #1651 

### What

* Adds CSS Variable parsing to the `StringInterpolator` as well as identifying that a CSS Variable is an `AnimatedString`.
* Adds tests based around ensuring that variables and their fallbacks can be correctly parsed.
* Updates documentation to reflect the new addition

### Checklist

- [x] Documentation updated
- [x] Demo added
- [x] Ready to be merged


docs/src/pages/basics.mdx
```diff
@@ -173,6 +173,7 @@ function Scrolling() {
 Springs take pretty much everything, they don't just handle numbers.
 
 - Colors (names, rgb, rgba, hsl, hsla, gradients)
+- CSS Variables and their fallbacks (as long as they're attached to the `window` object)
```
kindoflew, line 176: does this mean that vars not set on `#root` won't work?

packages/shared/src/createInterpolator.test.ts
```diff
@@ -167,4 +167,53 @@ describe('Interpolation', () => {
...
+        range: [0, 1],
+        output: ['var(--to)', 'var(--from)'],
+      })
+
+      expect(interpolation(0.0)).toBe('rgba(39, 126, 244, 1)')
+      expect(interpolation(0.5)).toBe('rgba(50, 191, 226, 1)')
+      expect(interpolation(1.0)).toBe('rgba(60, 255, 208, 1)')
+    })
+
+    it('should return the fallbacks if the variable cannot be found', () => {
```
kindoflew, line 208: should we have a test for fallbacks that are also CSS vars or is that overdoing it?

kindoflew (approved):
LGTM! Had two small questions in the comments, but they don't have to do with the functionality.

packages/shared/src/createInterpolator.test.ts
joshuaellis: No good idea 👍🏼 i'll add that.

docs/src/pages/basics.mdx
joshuaellis: AFAIK you have to set them like:

```css
:root {
  --from: #3cffd0;
  --to: #277ef4;
}
```

How i've written it is obviously a bit unclear, do you have an idea how it might be better written?

kindoflew: oh yeah, i meant `:root` 😅 . 

we could just replace '`window` object': "as long as they're declared on the `:root` pseudo-class."

or maybe: "CSS Variables (declared on `:root`) and their fallbacks."

MarkosKon:
Hi, I'm getting a "window is not defined" error on [`variableToRgba`](https://github.com/pmndrs/react-spring/blob/master/packages/shared/src/variableToRgba.ts#L21) during SSR (Gatsby site) on the `9.4.0` version. Is it acceptable to add a window guard (`typeof window === 'undefined'`)? Cool addition by the way.

joshuaellis:
Hey @MarkosKon can you open a new issue for this? 

This is something I must have forgotten & whilst I imagine you solution should work, I'm worried there might be a FOUC downside. If you want to make the PR infact, we can discuss there instead!

MarkosKon:
@joshuaellis I opened a new issue.

PR #1825, opened 2022-01-27
Fix missing commas in api object

chriscerie:
### Why

There are missing commas in the api object.

### What

There are now sufficient commas in the object.

### Checklist

- [x] Documentation updated
- [x] Ready to be merged


joshuaellis (approved):
Thanks for the contribution!

Issue #1799, opened 2021-12-22
[RFC][documentation] What do you think?


joshuaellis:
### Preface

Whilst I think our documentation coverage is _good_, I think it can be hard to find what you want. The information architecture of the site isn't quite right. There are pieces of documentation we want to be shouting about e.g. our [Imperative API](https://react-spring.io/common/imperatives-and-refs#imperative-api) but it's not really doing that.

### Going forward

I don't think it's something i'll be able to solve in a day or two, it'll be a classic gradual re-write / shape / tweak. But I welcome contribution either in code or suggestions here. All suggestions will be taken into account, but realistically we can't implement everyone's opinion. It'll be about trying to get the clearest message.

### How can you help?

To get me started, I need to understand the pain points! 
- What didn't you like about the documentation? 
- What did you find confusing or difficult using the library at first
- What do you like about our documentation compared to others?
- What other documentation sites do you like? (Doesn't have to be animation related)

### My own thoughts

For transparency here's a few thoughts i've had:

- We should have an `examples` page, to show case all of them, they're hiding with the related component!
- Search bar, it's 2021 [`algolia`](https://www.algolia.com/) is a thing, we should get on it!
- The basics are too long, they're not basics if they're that long!

krispya:
Hi Josh, overall the docs have served me well and only gotten better over time. There are two things that would suggest though:

1. With r3f being mixed with react-spring more often, it makes sense to have an imperative update guide. Maybe this is something I can help with?
2. Spring property controls on the Config page are useful to quickly visualizing spring motion without setting up my own controls. It would be great to enhance this to toggling visualizing with damping/frequency and tension/friction. Some of us are more used to seeing damping/frequency values. It would also be great to enhance this to visualize with a square moving from one side of the screen to the other. And if we make these enhancements I'd also suggest moving this to its own page section to help people test spring configs.

trm217:
Let me tell you about my first experience with the react-spring docs.

Around a year ago, I used/tried using react-spring for some micro-interactions in a page that I worked on. I was a bit overwhelmed by the docs at first, so I ended up copying examples onto the page. 
Once the example worked, I then tried to adapt the examples to my use case. This worked well for more primitive examples, but this didn't work at all for more complex examples with the useTransition hook.

[...]

PR #1835, opened 2022-02-13
docs: make links jump to beginning of the section

simlmx:
### Why

Currently going through the documentation is very annoying since it always bring you back to the big header. Many people have voiced that issue in #1799 

### What

Here I change the links in the table of content to point to the *section* (they all have a url pointing to them).
Note that I only "fixed" the links in the table of content, not the inline links of the doc.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

I'm not sure this is the proper approach but it was very straightforward to add.


orta:
I was just cloning the repo to make this PR, this should get merged - reading docs by having to scroll through a page of headers each time gets annoying pretty quickly

joshuaellis:
Looking at the deployment this doesn't work as intended, i think this is because the rehype/remark plug in doesn't use hyphens so `use-chain` should actually be `usechain`.

orta:
No worries, I'll send a new PR with fixed links 👍🏻

joshuaellis:
closing in favour of #1867

PR #1845, opened 2022-02-20
Docs fixes

bkrmadtya:
### Why

- Example code was not functioning as it was supposed to
- The instruction did not state clearly when the animation will be in effect which could be confusing for beginners (which was my case)

### What

Updated the documentation:
- Fixed the unused useSpring `styles` object that was not passed as style to the `animated.div`
- Clearly stated when the animation is shown

### Checklist

- [x] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

This was my first time trying `react-spring` and I tried the example in CodeSandBox but it was not working. Found why it was  not working and updated the documentation so, it will be easier for anyone who is trying this library for the first time.

**NB:** This is my first time contributing to any open source project.


joshuaellis:
Great catch! That's an issue with the docs package at the moment, so I need to take a look at that before I can merge this.

joshuaellis (approved):

PR #1849, opened 2022-02-25
fix: link color

Fasani:
### Why

Link looks bad on the first visit, but looks ok after `:visited` (likely why it was not spotted before), open in incognito mode to test, screenshots below.

Before: 
Screenshot 2022-02-25 at 16 55 45

After:
Screenshot 2022-02-25 at 17 04 34

### What

Small CSS change, unsure if there would be a preferred `:visited` colour and default colour?

### Checklist

- [ N/A ] Documentation updated
- [ N/A ] Demo added
- [ X ] Ready to be merged


joshuaellis (approved):

PR #1867, opened 2022-03-30
Links jump to section (part 2) + fix tarot images

orta:
### What

I went through all the links in #1835 and verified them against the site and also fixed the URLs to the tarot cards in the demos 

### Checklist

- [x] Documentation updated
- [x] Ready to be merged


orta:
I did a quick look into using rehype-slug which mdx uses to try and do this automatically, but it'd require pulling out the mdx file, parsing it and grabbing the header - which is a bit of an overkill

joshuaellis:
> I did a quick look into using rehype-slug which mdx uses to try and do this automatically, but it'd require pulling out the mdx file, parsing it and grabbing the header - which is a bit of an overkill

I agree, plus i'm working on redoing the documentation site, so this won't necessarily be an issue in the future.

joshuaellis (approved):

PR #1869, opened 2022-04-15
Add explicit semicolon to prevent syntax error in TypeScript 4.7

MichaelMitchell-at:
### Why

Without an explicit semicolon the TS parser treats the input as the new syntax, `typeof interpolator<Out>`
https://github.com/microsoft/TypeScript/issues/48711

### What

Add an explicit semicolon. I'm assuming that there is no auto-formatting step that will strip it away.

### Checklist

- [x] Ready to be merged


MichaelMitchell-at:
My company is still using react-spring v8. Any chance that this change could also be applied and released as v8.0.28?

MichaelMitchell-at:
TS will revert to old the old parsing behavior before 4.7 is released

PR #1876, opened 2022-04-25
Fix react-konva peerDependency

### LINKED ISSUE #1861: [bug]: React Spring cannot be installed with React 17 with strict peer dependency checking

- [ ] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [X] `@react-spring/konva`
- [ ] `@react-spring/zdog`

^9.4.4

### What's Wrong?

`@react-spring/konva` cannot be installed with `react@17` with strict peer dependency mode enabled.

`@react-spring/konva@9.4.4` requires `react-konva@^16.8.0  || ^17.0.0`. However, no package satisfying `react-konva@^16.8.0  || ^17.0.0` exists, because `react-konva` has not released an official `^17.0.0` yet, only pre-releases e.g. `17.0.2-5`. Installing `react-konva@^16.8.0`, of course, fails peer dependencies with `react@17`.

### To Reproduce

Make a new empty folder, add a package.json with:
```
{
    "name": "react-spring-test",
    "version": "0.0.1",
    "private": true,
    "devDependencies": {
        "@react-spring/konva": "^9.4.4",
        "react": "^17.0.0"
    }
}
```

and run `npm install --strict-peer-deps`

### Expected Behaviour

`npm install --strict-peer-deps` should succeed

### Link to repo

https://github.com/21echoes/react-spring-react-17-test

davbrito:
### Why

This PR resolves #1861 

As of now, the `@react-spring/konva` package has a peer dependency on `react-konva`:
https://github.com/pmndrs/react-spring/blob/a865bb7ba00d669d63483fdbe95339521a9e0ccb/targets/konva/package.json#L38

But react-konva uses a pre-release scheme on the v17. So the version used breaks peerDependencies of projects using react 17.

You can test it on the [npm semantic version calculator](https://semver.npmjs.com/):
image

### What

I added the version `^17.0.2-0` to the scheme so it could match for those versions:
image

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


joshuaellis:
Would this break if `react-konva` used something like `^17.0.2-1`? 🤔

davbrito:
> Would this break if `react-konva` used something like `^17.0.2-1`? 🤔

It would just not allow version `17.0.2-0`, so it shouldn't be harmful.

joshuaellis:
Hey I just saw you bumped the version number, is this because we need to use an explicit version number as `17.0.2-1` isn't obviously semantic?

joshuaellis:
> > Hey I just saw you bumped the version number, is this because we need to use an explicit version number as `17.0.2-1` isn't obviously semantic?
> 
> Yoy can see on the screenshots of the PR which versions would be allowed

right, i see, thanks! that wasn't super clear to me.

lavrton:
The proposed fix in the PR looks good to me. Thanks for pinging me, I was not aware current versions may produce such issues. I will think about better solution from `react-konva` side.

Issue #1794, opened 2021-12-21
[regression] transition.reverse is currently not possible


joshuaellis:
### Discussed in https://github.com/pmndrs/react-spring/discussions/1758

<div type='discussions-op-text'>

<sup>Originally posted by **ottotsuma** November 17, 2021</sup>
Hey all, 
I am trying to move to react-spring V9 from V8.

So in Spring V8 I use to do:

`bottomTransition.reverse()`

To get the animations to play in the reverse order do cards [1 ~ 5] would come out in the order 5, 4, 3, 2, 1. Under react-spring V9 I can no longer just .reverse() the reply because it's now a function call. So [1 ~ 5] becomes 1, 2, 3, 4, 5.

I tried the sort method found: [https://react-spring.io/hooks/use-transition#sort](https://react-spring.io/hooks/use-transition#sort)

However this just reverses the order of the cards in the deck not the order they come out. So [1 ~ 5] becomes [5 ~ 1] but now they come out 5, 4, 3, 2, 1 instead of 1, 2, 3, 4, 5.

I just want the animation order to be reversed.

```js
  const bottomTransition = useTransition(
    bottomTeamState,
    {
      from: { transform: "translate3d(-700%, 0, 0)" },
      enter: { transform: "translate3d(0%, 0, 0)" },
      leave: { transform: "translate3d(700%, 0, 0)" },
      trail: 175,
      config: {clamp:   true, mass: 3, tension: 150, friction: 38 },
  });

bottomTransition(( style, item, transition, index ) => {
```</div>

sagarchoudhary96:
@joshuaellis i would like to work on this one. Would be great to know a little bit more about starting point to look for the fix for this as this is first time i'm working with react-spring.

sagarchoudhary96:
@joshuaellis how do i make changes locally to this repo and test out in my local project?

joshuaellis:
Hey @sagarchoudhary96, you'll want to look at the useTransition hook and towards the bottom where we handle the users `render` function, I haven't given it much thought from that.

Working on the repo, you should fork the repo and work there, there's a demo folder in the repo where you'll see all the sandboxes. You should probably just change an existing `useTransition` example if possible or create a very simple and functional demo to work with whilst implementing the changes.

Finishing up I don't think we need a dedicated example of the prop, but we should add a test to cover this regression of the public API.

sagarchoudhary96:
[...]

PR #1896, opened 2022-05-19
toggle reverse on useTransitions

Supersteve128:
### Why

It is not currently possible to reverse the order of the useTransition animation. 

### What

I changed the render transition function to allow an optional reverse prop and added the optional toggleReverse prop in the hook to change the transition ordering after every render.

### Checklist

- [ ] Documentation updated
- [x] Demo added
- [x] Ready to be merged

This is an attempt to solve [https://github.com/pmndrs/react-spring/issues/1794](url)


joshuaellis (changes requested):
It looks like you've gone for two approaches here, i'd prefer to go for the prop styled method as opposed to adding another argument to `transition`.

I think `reverse` would be better because this is in sync with the language we use in regards to `loop`.

We should also add some tests for this new feature.

misbahkhalilaz (approved):

PR #1897, opened 2022-05-20
fix: Make decay easing respect the `config.precision` prop

claytercek:
### Why

The current behavior for a decay animation is to finish when the difference between the current spring value and the value during the previous frame is less than the arbitrary value of `0.1`. In most cases this is alright, but if the scale of the animation is significantly small, then we start to notice some weird behavior, where the animation seems to stop prematurely. Here's an example codesandbox where you can simulate different scales: https://codesandbox.io/s/react-spring-decay-scale-j5kfqm?file=/src/App.js

One situation where this is particularly apparent is when using react-spring in combination with react-three-fiber, and animating in world space. Demo here: https://codesandbox.io/s/r3f-inertia-camera-qqhvxk?file=/src/InertiaCamera.js 

### What

This PR replaces that arbitrary `1.0` value with the same precision value being used for spring easing. 

I've moved the precision calculation to a slightly higher scope, so that it can be shared between both spring and decay easing. 

### Checklist

- [x] Documentation updated
- [x] Ready to be merged


packages/core/src/SpringValue.test.ts
```diff
@@ -467,6 +467,35 @@ function describeConfigProp() {
         expect(global.countBounces(spring)).toBeGreaterThan(0)
       })
     })
+    describe('the "precision" prop', () => {
+      describe('stops decay animation when step difference is less than precision', () => {
```
joshuaellis, line 471: Would you be able to add a test around it effecting regular springs without `decay` applied?

joshuaellis (changes requested):
Not sure how I missed this so sorry for the late reply, but this is a great catch thank you! I have one small request though, see comment!

packages/core/src/SpringValue.test.ts
claytercek: Yep! Added a test for spring easing in commit cd43a70

joshuaellis (approved):

PR #1925, opened 2022-07-04
fix: make children on AnimatedComponent optional

LinusU:
### Why

It fixes the issue described here: https://github.com/pmndrs/react-spring/pull/1705/files#r913068712

### What

I have made the `children` property optional on the `AnimatedComponent` component.

I tested this out in my project by editing the file in `node_modules`, making the same change as proposed here.

### Checklist

- [x] Ready to be merged


joshuaellis:
Thanks for this! I've ran the tests so we can see if it's working fine.

joshuaellis (approved):

PR #1938, opened 2022-07-16
chore: add react peer dependencies

CharlesStover:
All the packages in this repo require `react` and `react-dom` as peer dependencies, and `react-spring` does not provide them even though they are listed as dependencies. It needs to "pass the buck" to the consumer of `react-spring` in order to get rid of these warnings on install:

```
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (p45d7f), requested by @react-spring/core
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (p89414), requested by @react-spring/konva
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (pad289), requested by @react-spring/native
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (p67454), requested by @react-spring/three
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (p313ce), requested by @react-spring/web
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react (p43935), requested by @react-spring/zdog
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react-dom (pcccdb), requested by @react-spring/web
➤ YN0002: │ react-spring@npm:9.4.5 doesn't provide react-dom (p0095e), requested by @react-spring/zdog
```

This fix is validated by adding the following to `.yarnrc`:

```yaml
packageExtensions:
  react-spring@*:
    peerDependencies:
      react: '*'
      react-dom: '*'
```

But I'd rather not do the YAML implementation because it's dead code if this fix is ever applied to the source, and other devs on the team won't understand maintaining this. I'd rather it be fixed at the source, hence the PR. :)


packages/react-spring/package.json
```diff
@@ -33,5 +33,9 @@
     "@react-spring/three": "~9.5.0",
     "@react-spring/web": "~9.5.0",
     "@react-spring/zdog": "~9.5.0"
+  },
+  "peerDependencies": {
+    "react": "*",
```
joshuaellis, line 38: this peer dep is incorrect, I would look at the packages use of it and reference that.

joshuaellis (changes requested):
See comments

CharlesStover:
I've updated it with the versions from `@react-spring/core`.

CharlesStover:
@joshuaellis Sorry to ping. It's been a few weeks. The changes were made. Is it good to 🐿?

joshuaellis:
No totally fair to ping me. I think you need to run `yarn` and push the new lockfile and then all three checks _should_ pass 👍🏼 I can then double check tomorrow if they pass.

CharlesStover:
@joshuaellis Sorry for the delay. I've updated the lock file.

joshuaellis (approved):

eugenet8k:
Thank you for the change. But it seems the values in peer dependencies are not correct. Now it is:

```
  "peerDependencies": {
     "react": "^16.8.0  || >=17.0.0 || >=18.0.0",
     "react-dom": "^16.8.0  || >=17.0.0 || >=18.0.0"
   }
```

but having `>=` twice does not make sense, if the version bigger than 18 it also will be bigger than 17. This online check tool demonstrates this

https://jubianchi.github.io/semver-check/#/^16.8.0%20%20||%20%3E%3D17.0.0%20||%20%3E%3D18.0.0/20

Did you perhaps want to have this syntax instead?

```
  "peerDependencies": {
     "react": "^16.8.0  || ^17.0.0 || ^18.0.0",
     "react-dom": "^16.8.0  || ^17.0.0 || ^18.0.0"
   }
```

this is more common for libraries as it allows to have any version within the major version range, but it won't apply to React v19. Your current version range would match any React blindnessly, even React v100.

joshuaellis:
This isn't released yet, feel free to make a PR and submit your findings so we can discuss separately.

eugenet8k:
@joshuaellis I made a PR that failed checks:

```
➤ YN0028: │ -    react: ^16.11.0  || >=17.0.0 || >=18.0.0
➤ YN0028: │ +    react: ^16.8.0 || ^17.0.0 || ^18.0.0
➤ YN0000: │      three: ">=0.126"
➤ YN0000: │    languageName: unknown
➤ YN0000: │    linkType: soft
➤ YN0000: │  
➤ YN0000: │ 
➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
➤ YN0000: └ Completed
```

It seems I am not allowed to modify yarn.lock file in my commit, but how to address this change in peerDependencies then?

joshuaellis:
@eugenet8k youd have to run yarn to modify the lock file which is fine because you're changing peer deps.

eugenet8k:
@joshuaellis it seems I got successful checks this time, please review and thank you for all the work with this repo.

PR #1969, opened 2022-09-01
Use proper peer dependencies range condition

eugenet8k:
* It targets React v16, v17, and v18, and only those

fixes #1938


joshuaellis:
if you run `yarn` it should update the `lockfile` correctly for you.

PR #1984, opened 2022-09-20
docs: added examples for interpolation function 'to'

anshullaikar:
### Why

The docs for [interpolation](https://react-spring.dev/common/interpolation#interpolations) are missing the `to` function implementation. I had to go through some examples to figure out how to use the `to` function. My addition to the docs should help other people who are facing a similar issue.

### What

Have added an example of an implementation of `to` which helps explain it's use case with `useTransition` as well.  Please let me know if any changes are required in the formatting.

### Checklist

- [x] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/docs/src/pages/common/interpolation.mdx
+++ b/docs/src/pages/common/interpolation.mdx
@@ -13,6 +13,70 @@
 | output           | undefined | Array of mapped output ranges                    |
 | map              | undefined | Value filter applied to input value              |
 
+## Example 
+
 A range shortcut: `value.to([...inputRange], [...outputRange])`
 
 Or a function: `value.to(v => ...)`
+
+## Alternate method using `to` function
+#### For single variables
+
+Example is using `to` for interpolating `y` style for a `transform: translate3d(x,y,z)` animation
+
+```
+import { useTransition, to } from "react-spring";
+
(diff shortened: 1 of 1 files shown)


anshullaikar:
@joshuaellis Any comments?

anshullaikar:
@joshuaellis any comments?

anshullaikar:
Have updated the commit message and PR title as per the Conventional Commit guidelines

joshuaellis (changes requested):
These examples are overly complicated e.g. why are are mapping `elements` in this example? Then you just pass "settings" it's all too vague as well.

In the first example i'd argue the use of the `to` function is superfluous when you can just do `styles.y.to` and then you're passing the `key` object which you don't need to do so that's incorrect.

IMO these examples should be thorough and simple.

joshuaellis:
Closing this because i've added some documentation around this on the new doc site – https://beta.react-spring.dev/docs/advanced/interpolation

anshullaikar:
@joshuaellis Wow, the new documentation site looks great! And thanks for the feedback, will add simpler examples if I think of any other documentation PRs :slightly_smiling_face:

Issue #1725, opened 2021-10-25
api.set does not work for useSprings hook


fast-reflexes:
## Intro

When doing animations, we sometimes have the need to set a value without animating it. An example is when showing a tooltip when hovering over a certain area of the screen; as long as the mouse is in this area, we want to animate the positional changes of the tooltip but if we move out of the area from the left, and then enter the area from the right, we want to position the tooltip at the place of entrance immediately, while at the same time setting it to be visible. I was happy to find the complementary `api.set` and `api.start` for these occasions.

## 🐛 Bug Report

The above behaviour works great for `useSpring`, but not for `useSprings`. On https://react-spring.io/hooks/use-springs there is the following example:

    api.start(index => ({ opacity: 0 }))

where you think it's possible to use the api returned by `useSprings` to make individual changes to the returned springs. This is also the case, but only for the `api.start` method, not for the `api.set` method. As I do the same for set:

    api.set(index => ({ opacity: (100 - index) / 100 }))

the editor complains and the expected effect does not happen. The same kind of change works great with `useSpring`.

## To Reproduce

Check https://codesandbox.io/s/react-spring-usesprings-bug-z2hic and follow the instructions there

Simply use `useSprings` and try to run `api.set` for individual springs on its returned api and the set effect will not take place.

## Expected behavior

I expect the behaviour of `useSpring` to be the exact same in `useSprings` but generalized over multiple springs. I also expect the `api.start` method of `useSprings` to be similar to `api.set` which it does not appear to be.

All in all, this means that I expect `api.set` to work and that it should have the same effect as `api.start` but without an animation, but instead an immediate resulting effect.

## Link to repro (highly encouraged)

https://codesandbox.io/s/react-spring-usesprings-bug-z2hic

## Environment

- `react-spring` v9.3.0
- `react` v17.0.2

Thanks for a GREAT library! 🌹

alexwidua:
Not sure if it's a bug or intented, but to anyone running into the same 'problem': 
Instead of trying to use `api.set(...)`, return an `immediate: true` prop value in addition to your spring values.

ex.
`api.start((i) => {
	// const x = ...
	return { x, immediate: true }
})`

uelibaumgartner:
@alexwidua I have the same issue... unfortunately if I use` api.start` I guet a weird glitch where I see one frame of the element moving, which I don't get with `api.set`.

Would be great if some one could implement the missing piece of the api

eden-lane:
[...]

PR #1990, opened 2022-09-29
Update SpringRef.ts

AndriesVDW:
### Why

Resolves https://github.com/pmndrs/react-spring/issues/1725

### What

I made it possible to use it the same way we can use start and update by passing a callbacl function that loops through the controllers.

### Checklist

- [x ] Ready to be merged


joshuaellis:
Thanks for this contribution, I'm wondering if we should be extending this to the other methods – `pause`, `resume` & `stop`? Feels like this would be a good opportunity to do this. I'm also wondering if this would be a good opportunity to add tests for `SpringRef` 🤔

PR #1999, opened 2022-10-15
docs: Update preset section on spring config

heyfirst:
### Why
I see a lot of value from [this preset section](https://react-spring.dev/common/configs#presets) in old docs. I believe it would be nice if we can have similar thing in new docs as well.

### What
- Move preset section to be under the config visualization
- Add preset visualization with the same idea of [original preset section](https://react-spring.dev/common/configs#presets)

### Checklist
- [x] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### Comments
Here is my first time coding in remix and `jsx` tag on MDX. It is really cool that we can define a code on MDX and automatically run on codesandbox. That's brilliant!

### DIFF
--- a/docs/app/routes/docs/advanced/config.mdx
+++ b/docs/app/routes/docs/advanced/config.mdx
@@ -116,6 +116,60 @@ const MyComponent = () => {
 }
 ```
 
+## Config Visualizer
+
+```jsx live=true template=configPlayground showCode=false
+import { useRef, useState } from 'react'
+import { useSpring, animated } from '@react-spring/web'
+import { useControls } from 'leva'
+
+export default function Card() {
+  const cardRef = useRef(null)
+  const config = useControls({
+    mass: 1,
+    tension: 170,
+    friction: 26,
+    clamp: false,
+    precision: 0.01,
+    velocity: 0,
+  })
+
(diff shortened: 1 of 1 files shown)


docs/app/routes/docs/advanced/config.mdx
```diff
@@ -116,6 +116,60 @@ const MyComponent = () => {
 }
 ```
 
+## Config Visualizer
+
+```jsx live=true template=configPlayground showCode=false
+import { useRef, useState } from 'react'
```
joshuaellis, line 122: useState isn't being used, I think we should remove this.

joshuaellis (changes requested):
Small change, i'm wondering if we should combine the preset visualiser and the config editor, do you see any value in keeping them separate?

heyfirst:
@joshuaellis Good point! I didn't see any value to separate it either. I will update to add one more preset option to the config visualizer above. Will ping you once it's ready to review :)

heyfirst:
Sorry for super-long silence, I have been into some initiatives on my current work, many things to catchup, I will update this PR on this weekend.

joshuaellis:
closing due to inactivity

PR #2002, opened 2022-10-22
feat(docs): add r3f doc guide

joshuaellis:
### What

* Adds an `r3f` guide to `react-spring` documentation


joshuaellis:
Be great to get thoughts and opinions from you both.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+    </Canvas>
+  )
+}
+```
+
+## Use the imperative API
+
+In the example below, we use the power of the `react-spring`'s api to lean into the imperative requirements
+of working with performant webGL scenes. The blob follows you round the canvas (lines 70-82) & scales on
+interaction (lines 46-56) without a single react render to cause these updates.
```
CodyJasonBennett, line 86: Is it possible to make `lines 70-82` and `lines 46-56` linkable?

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+
+## Why use the library?
+
+A common question asked is why use `react-spring` with `react-three-fiber` when you can use the `useFrame` hook to update your meshes & objects
+instead without knowing another API. Well this is a great question, lets consider a simple use-case.
+
+I have a distortion blob and I want to it to change color on click. You could do this with `useFrame` to perform frame by frame updates, `useRef`
+to access the material object and use the `THREE.Color.lerp` function, slowly incrementing by lerp value until the color is reached. But this then
+requires that I keep track of a `THREE` instance of `Color` which _can_ lead to memory leaks if not handled correctly and not only that, but it would
+be a lot of code. Then you'd need to think about writing your own easing functions. With `react-spring` you can do this in a few lines of code.
```
CodyJasonBennett, line 40: I think it's important to mention that react-spring's animations are physically correct. It is critical of motion design for animations to look realistic; otherwise, it's fake and awkward -- look at Dribbble to see what I mean. That is not to say that linear easings don't have their place; they are more appropriate for reactive elements or interactions where a response must be immediate, but that is inbuilt into CSS and `Vector3#lerp`. Furthermore, the flexibility to start/stop and replay animations, particularly with state and device motion preferences, makes this a uniquely accessible library from both a DX and UX perspective. These are points I'd highlight upfront since no other library has come close to offering the above.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+
+In addition we use the provide a function to the `config` prop instead of an object to deliver a more sticky
+config for the spring in general, but to give the blob a bouncy feel when the `scale` key is changed – lines 18-30.
+
+Finally, because we're using the position of the mouse which can be considered a `Vector2` and the `position`
+of a `mesh` is a `Vector3` we use a custom interpolation via the `to` method of a `SpringValue` to interpolate
+the array, this can be seen on line 97.
+
+Why not read more about it the [imperative](/docs/concepts/imperative-api) and [interpolation](/docs/advanced/interpolation) APIs?
+
```
CodyJasonBennett, line 96: You might want to link related readings with a subtle aside or note. In readmes like [pmndrs/react-ogl](https://github.com/pmndrs/react-ogl) I'd say something like (also see ...) so it doesn't break from the content flow.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+    api.start({
+      position: [0, 0],
+    })
+  }, [])
+
+  const handlePointerMove = useCallback(
+    e => {
+      if (isOver.current) {
+        const x = (e.offsetX / width) * 2 - 1
+        const y = (e.offsetY / height) * -2 + 1
```
CodyJasonBennett, line 171: I believe this NDC math is inbuilt into R3F's state model as `pointer`. I'd have to defer to @drcmda as for surrounding patterns but you could greatly reduce boilerplate with `<Canvas onCreated={(state) => state.events.connect(window)} />`.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+might find toward the end of an animation that there's a subtle jump,
+which is visible in [this demo](https://y18m1.csb.app/). It's not pretty, is it?
+
+Whilst by default it would be nice to have this issue resolved without you having
+to interact and this is something we'll consider for the next `breaking change`
+in the meantime what you can use is the [`precision` config prop](/common/configs#configs) to avoid this.
+
+By setting the prop to a value like `0.0001` you can notice there is no jump towards 0.
+This is because the precision prop is used to figure out how close the animated value
+can get to the end goal before we consider the animated value to be equal to the end goal.
```
CodyJasonBennett, line 401: You'll want to give a code example or make this interactive for those referring to this guide for the troubleshooting section.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+import { Canvas } from '@react-three/fiber'
+
+const AnimatedMeshDistortMaterial = animated(MeshDistortMaterial)
+
+const MyScene = () => {
+  const [clicked, setClicked] = useState(false)
+
+  const springs = useSpring({
+    color: clicked ? '#569AFF' : '#ff6d6d',
+  })
```
CodyJasonBennett, line 55: Later, we'll see `useSpring` accept a function to return an `api` object. It may be worth noting its overloads after this section.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+meta:
+  title: R3F | React Spring
+  'og:title': R3F | React Spring
+  'twitter:title': R3F | React Spring
+  description: A detailed guide to using React Spring with R3F.
+  'og:description': A detailed guide to using React Spring with R3F.
+  'twitter:description': A detailed guide to using React Spring with R3F.
+  'og:url': https://beta.react-spring.dev/docs/guides/react-three-fiber
+  'twitter:url': https://beta.react-spring.dev/docs/guides/react-three-fiber
+sidebar_position: 1
```
CodyJasonBennett, line 11: Is there a way to add tags or meta information so this is searchable under "R3F", "react-three-fiber", "three", etc.? You don't want the article title to not have searchable keywords on the site.

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+---
+
+# React Three Fiber
+
+:::note
+It's assumed you have a base understanding of `react-spring` and `react-three-fiber`. If you're new to
+either, check out our [getting started](/docs/getting-started) or alternatively, the [r3f](https://docs.pmnd.rs/react-three-fiber) docs.
+:::
```
CodyJasonBennett, line 19: Nit, you'll want to consistently spell or refer to R3F or react-three-fiber here. @hmans has asked before about how to refer to R3F as a package -- whether that's `@react-three/fiber`, react-three-fiber, R3F, etc.

docs/app/routes/docs/guides/react-three-fiber.mdx
krispya: Just want to second, true spring motion is a huge plus. Even damping is an approximation. The ability to design physically accurate, fully interruptible animations, especially for 3D user interactions, is necessary for a polished feel in an app. This video might give you some inspiration, the visuals are great: https://youtu.be/KPoeNZZ6H4s?t=195

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+## Working with spring events
+
+Sometimes, it's necessary to sync the state of a spring with the an external source. This can be done with the [event system](/docs/advanced/events)
+built into react-spring.
+
+Take the following example, we have multiple blobs on our screen that start in different places and a component higher in our scene graph needs to to know
+the position of each blob. Because the `position` is controlled by `useSpring` you can't simple submit `springs.position` to the store because you'll
+be dispatching the whole `SpringValue` object, which is unnecessary and can weigh down your external store.
+
+Instead, you can use the `onChange` event handler to `get` the value of your springs and react to them accordingly. The code below is a conveluted example
```
krispya, line 230:
```suggestion
Instead, you can use the `onChange` event handler to `get` the value of your springs and react to them accordingly. The code below is a convoluted example
```

docs/app/routes/docs/guides/react-three-fiber.mdx
krispya: Another bonus that came to mind is framerate independence. You can get the same with damping functions, but then damping is much harder to design with since it has fixed dynamics (since all damping is doing is approximating spring motion).

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+interaction (lines 46-56) without a single react render to cause these updates.
+
+In addition we use the provide a function to the `config` prop instead of an object to deliver a more sticky
+config for the spring in general, but to give the blob a bouncy feel when the `scale` key is changed – lines 18-30.
+
+Finally, because we're using the position of the mouse which can be considered a `Vector2` and the `position`
+of a `mesh` is a `Vector3` we use a custom interpolation via the `to` method of a `SpringValue` to interpolate
+the array, this can be seen on line 97.
+
+Why not read more about it the [imperative](/docs/concepts/imperative-api) and [interpolation](/docs/advanced/interpolation) APIs?
```
krispya, line 95:
```suggestion
Why not read more about the [imperative](/docs/concepts/imperative-api) and [interpolation](/docs/advanced/interpolation) APIs?
```

docs/app/routes/docs/guides/react-three-fiber.mdx
```diff
@@ -0,0 +1,401 @@
...
+    <Canvas>
+      <ambientLight intensity={0.8} />
+      <pointLight intensity={1} position={[0, 6, 0]} />
+      <MyScene />
+    </Canvas>
+  )
+}
+```
+
+## Working with spring events
```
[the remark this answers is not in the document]
krispya: Maybe a good syncing example is syncing to an `Object3D` that is not represented in JSX. The `animated` examples wouldn't be able to do this and it doesn't require any complicated application design to demonstrate the strategy.

docs/app/routes/docs/guides/react-three-fiber.mdx
joshuaellis: The search is DocSearch by algolia, I think (when the crawler is ran) it should be searchable by _lots_ of things.

docs/app/routes/docs/guides/react-three-fiber.mdx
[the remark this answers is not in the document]
joshuaellis: > Maybe a good syncing example is syncing to an Object3D that is not represented in JSX.

Do you have an example that comes to mind? 🤔 I'm totally up for simplifying this.

docs/app/routes/docs/guides/react-three-fiber.mdx
joshuaellis: Really good point.

joshuaellis:
I haven't replied to every comment, but overall, thank you both very much for your feedbacks. I won't be re-asking you to review (don't want to pest) but until the PR is merged / comments are resolved, please feel free to comment adhoc.

docs/app/routes/docs/guides/react-three-fiber.mdx
joshuaellis: fantastic points 🙇🏼‍♀️

docs/app/routes/docs/guides/react-three-fiber.mdx
joshuaellis: No unfortunately not, the best i could do is highlight them.

docs/app/routes/docs/guides/react-three-fiber.mdx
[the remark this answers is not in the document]
krispya: It mostly gets used for more complicated workflows where you want an `Object3D` proxy that doesn't have its matrix auto-updated because it is on the scene graph. For example, this is done in the Drei `Instance` component to calculate things like position before feeding it into the buffer array. I'm not sure offhand what the best condensed version of this would be, maybe something to do with instancing or a simple particle system where you are feeding the resulting matrices into a shader.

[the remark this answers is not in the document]
joshuaellis: That could be a great example to make. 
I think for the purpose of the doc i'll keep as is because its narrative is you're building up a thing over time.

Thanks for the thoughts!

PR #2054, opened 2023-01-17
docs(fix): 'Learn more about targets' href

prichodko:


prichodko (commented):
Fix incorrect targets URL on the home page.

joshuaellis (approved):
Thanks!

PR #2056, opened 2023-01-19
docs(fix): cannot animate between AnimatedString and AnimatedValue, as the "to" prop suggests

Jood80:
### Why
I ran to this error today while I was reading the [animated components section](https://www.react-spring.dev/docs/concepts/animated-elements#animating-components).  The example I was trying to implement is the one attached in the code below.
I'm using Next.js and the error has been resolved once I updated the type of `to.y`: `string`, instead of  `to.y`: `number`.  

<div  align="left" dir="auto">
  animated-elements
    animated-elements  
</div>

### What
- Change the type of `to.y` coordinate to match the one used in `from.y`; which is happen to be a string in this example.

### Checklist
 - [x] Documentation updated
 - [ ] Demo added
 - [x] Ready to be merged

### DIFF
--- a/docs/app/routes/docs/concepts/animated-elements.mdx
+++ b/docs/app/routes/docs/concepts/animated-elements.mdx
@@ -82,7 +82,7 @@ const App = () => {
     },
     to: {
       opacity: 1,
-      y: 0,
+      y: '0',
     },
   })
 


joshuaellis (approved):
Great typo catch! Thank you 🌟

PR #2091, opened 2023-02-13
fix: allow container ref with null and undefined for useResize

floriangosse:
### Why

Allow to pass nullish container references to `useResize`.

### What

Extended type of `useResize` to allow `null` and `undefined` as reference value types as well.

### Checklist

- [x] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


joshuaellis (approved):
Nice addition! Thank you 🙇🏼‍♀️

Issue #2097, opened 2023-02-24
[bug]: ESM failed to resolve with Webpack


zkylearner:
### Which react-spring target are you using?

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

9.7.0

### What's Wrong?

ERROR in ./node_modules/@react-spring/core/dist/index.mjs 1596:0-45

Module not found: Error: Can't resolve '@react-spring/types/animated' in '.../my-app/node_modules/@react-spring/core/dist'
Did you mean 'animated.js'?
BREAKING CHANGE: The request '@react-spring/types/animated' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

### To Reproduce

Code running locally has an error

### Expected Behaviour

no error

### Link to repo

https://codesandbox.io/s/bold-mahavira-y26di2?file=/src/App.js

joshuaellis:
I cant see an error in the reproduction?

LoicHuart:
I have the same error on my project.
It's a react app with typescript.
And I use craco to create aliases and be able to import my files with @/.
it may be due to this configuration, I'm not sure.


```
// craco.config.js
const path = require('path');

module.exports = {
    webpack: {
        alias: {
            '@': path.resolve(__dirname, './src'),
        },
    },
};
```

GivenCui:
I also face this error today. Before that, it was working fine.

joshuaellis:
Please stop saying you're getting the error and not providing a reproduction – it's really unhelpful and doesn't make anything happen quicker.

nongtiny:
> Please stop saying you're getting the error and not providing a reproduction – it's really unhelpful and doesn't make anything happen quicker.

@joshuaellis 

I'm sorry. What kind of reproduction do you want to see?

For the test URL, I can't build the reproduction for you because When I do `react-scripts build,`  it has this error also. Build is failed.

I'm not sure if this can help you, but the `@react-spring/types` has an update last 13 hours ago. I don't know that this might be the case or not. ref: https://www.npmjs.com/package/@react-spring/types?activeTab=versions

[...]

PR #2099, opened 2023-02-25
fix: specify extension for ESM export

chenxsan:
### Why

Closes https://github.com/pmndrs/react-spring/issues/2097 due to https://github.com/webpack/webpack.js.org/issues/3975

### What

Just specify extension for those `export`.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

### DIFF
--- a/.changeset/seven-eyes-float.md
+++ b/.changeset/seven-eyes-float.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


joshuaellis:
I'll keep this open for the time being, but like I said in the issue I want to look further into the export system we have specifically looking at libraries like react-aria and probably redux-toolkit@2.

Thanks for this

Issue #2146, opened 2023-04-14
[bug]: useSpring Events not firing (Next.js 13)


lairdkruger:
### Which react-spring target are you using?

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

9.7.2

### What's Wrong?

API events are not triggering at all.

### To Reproduce

Setup a Next.js 13.3.0 project, try useSpring with events. Try to do something in the event eg: 
` const [{ x }, api] = useSpring(
    () => ({
      x: 0,
      // onChange is currently broken with Next App structure
      onChange: {
        x: () => {
          console.log("onChange"); // This should be logging the value on drag
        },
      },
    }),
    []
  );`

### Expected Behaviour

Expected console to log "onChange".

### Link to repo

https://codesandbox.io/p/sandbox/dawn-leaf-lqof4b?selection=%5B%7B%22endColumn%22%3A11%2C%22endLineNumber%22%3A12%2C%22startColumn%22%3A11%2C%22startLineNumber%22%3A12%7D%5D&file=%2Fpages%2Fworking%2Findex.tsx

bramarcade:
I'm also encountering this bug, with @react-spring/web@9.6.1 and next@13.4.1, using the app router as in the above repo. The springs work just fine, but events like onStart do not fire.

ionutpopa:
Yup, onStart, onRest do not work, as well as things like api.start or api.stop.

skibitskiy:
I found that something happens with the React life-cycle within `useSprings`. There is `useIsomorphicLayoutEffect`, that should run with [updates](https://github.com/pmndrs/react-spring/blob/bbfc8be61648122c3a3526e25534d10911c81db4/packages/core/src/hooks/useSprings.ts#L204) array, that consist of config, callback and other props. `Updates` is simple array variable (not react ref) and it rewrites on every render, but it calculates in useMemo on first render. So, I debugged on some examples on react-spring site and found that `useIsomorphicLayoutEffect` should be called with not empty updates array, but in Next.JS 13 effect calls with empty array

skibitskiy:
[...]

PR #2207, opened 2023-10-01
fix: save updates between renders in useSpring in Next.JS 13

skibitskiy:
### Why

It's a cheap fix, maybe should correctly configure Next.js
I found that something strange happens with the React life-cycle within useSprings in the Next.js 13. There is useIsomorphicLayoutEffect, that should run with [updates](https://github.com/pmndrs/react-spring/blob/bbfc8be61648122c3a3526e25534d10911c81db4/packages/core/src/hooks/useSprings.ts#L204) array, that consist of config, callback and other props. Updates is simple array variable (not react ref) and it rewrites on every render, but it calculates in useMemo on first render. So, I debugged on some examples on react-spring site and found that effect should be called with not empty updates array, but in Next.JS 13 effect calls with empty array

resolves: https://github.com/pmndrs/react-spring/issues/2146

### What

I wrapped updates variable into react ref

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useSprings.ts
+++ b/packages/core/src/hooks/useSprings.ts
@@ -128,7 +128,7 @@ export function useSprings(
   )
 
   const ctrls = useRef([...state.ctrls])
-  const updatesRef = useRef([])
+  const updatesRef = useRef<any[]>([])
   const { current: updates } = updatesRef
 
   // Cache old controllers to dispose in the commit phase.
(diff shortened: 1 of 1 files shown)


skibitskiy:
Sorry, it seems that this problem caused by React 18 strict mode in the dev mode. Solution – is turn off strict mode in the config

https://github.com/pmndrs/react-spring/issues/2146#issuecomment-1743513157

PR #2222, opened 2023-11-20
Adding bouncing ball react-spring example

Misterlinux:
### Why

Adding a demo for the react-spring controller section

### What

Added a Codesandbox Demo to the react-spring examples.

### Checklist

- [ ] Documentation updated
- [X] Demo added
- [ ] Ready to be merged

Feel free to message me in case any edit is needed to my code


joshuaellis:
This example appears to use non-english variable names and uses bootstrap-css which isn't something we really feel the need to advocate in addition it's not up to a standard I would expect e.g. not using `jsx` files and using `@react-spring/animated` directly.

PR #2244, opened 2024-01-05
fix update useSprings link in useSpring to /docs/components/use-springs

olaracode:
### Why

There is an error in the documentation with the useSpring page specifically.

When the useSprings hook is mentioned the link is giving a 404 since it is redirecting to .../use-use-springs

### What

Even tho I didn't see much difference between the other usage of the md link in the other pages of the docs. I did notice that the route start from /docs/components/ in other instances and in this one it started since /component/... 

So I simply added the /docs/ to the md

### Checklist

- [X] Documentation updated
- [ ] Demo added (Not really sure how to provide a demo)
- [ ] Ready to be merged

### DIFF
--- a/docs/app/routes/docs/components/use-spring.mdx
+++ b/docs/app/routes/docs/components/use-spring.mdx
@@ -13,7 +13,7 @@ sidebar_position: 1
 
 # useSpring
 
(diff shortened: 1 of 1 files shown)


joshuaellis:
duplicate of #2256

PR #2257, opened 2024-02-26
docs: modified useScroll usage code

Gaic4o:
### Why

Upon attempting to use the useResize hook as illustrated in the current documentation, I encountered an error stating that an argument was expected by useResize, but none was provided. This was specifically related to the initial usage example

```jsx
const { width, height } = useResize();
return <animated.div style={{ width, height }}>Hello World</animated.div>
```

This led to some confusion, as it was not immediately clear that useResize requires at least one argument to function correctly.

### What

To make it easier for people who use the useResize Hook to understand, we have modified the Usage example code.

### Checklist

- [x] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

If there are any changes I need to make to my code, I would appreciate your feedback.

### DIFF
--- a/docs/app/routes/docs/utilities/use-resize.mdx
+++ b/docs/app/routes/docs/utilities/use-resize.mdx
@@ -21,12 +21,30 @@ to the window, by passing a `container` you can observe that element's size inst
 ## Usage
 
 ```tsx
+import { useRef } from 'react'
 import { useResize, animated } from '@react-spring/web'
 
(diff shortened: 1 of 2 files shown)


joshuaellis:
I find the example you've provided more unclear, perhaps you can remove the noise around it e.g. background color, interpolating etc. to make it much more bare bones and simpler to understand the minimum required to use the hook?

PR #2281, opened 2024-05-25
Remove Optional Chaining Operator

crazyair:
### Why

Optional chains may cause errors in some lower version environments

### Checklist

- [x] Documentation updated
- [x] Demo added
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -205,14 +205,14 @@ export function isAsyncTo(to: any) {
 
 /** Detach `ctrl` from `ctrl.ref` and (optionally) the given `ref` */
 export function detachRefs(ctrl: Controller, ref?: SpringRef) {
-  ctrl.ref?.delete(ctrl)
-  ref?.delete(ctrl)
+  ctrl.ref && ctrl.ref.delete(ctrl)
+  ref && ref.delete(ctrl)
 }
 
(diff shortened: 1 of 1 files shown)


crazyair:
The problem is that this brought in https://github.com/pmndrs/react-spring/compare/v9.6.1...v9.7.0

joshuaellis:
Optional chaining is accessible in 94% of browsers – https://caniuse.com/?search=optional%20chaining I don't think we need to omit it, you should transpile your code to the correct target should you need to target a browser that does not support it.

crazyair:
> Optional chaining is accessible in 94% of browsers – https://caniuse.com/?search=optional%20chaining I don't think we need to omit it, you should transpile your code to the correct target should you need to target a browser that does not support it.

But it was good before 9.6.1.

crazyair:
https://npmmirror.com/package/@react-spring/core/files/dist/react-spring-core.cjs.dev.js?version=9.6.1#L157

image


https://npmmirror.com/package/@react-spring/core/files/dist/react-spring_core.modern.development.mjs?version=9.7.3#L124
image

crazyair:
https://npmmirror.com/package/antd/files/es/affix/index.js?version=5.17.3#L88-L89

image

https://github.com/ant-design/ant-design/blob/master/components/affix/index.tsx#L151
image

crazyair:
@joshuaellis Modify the compiled script?

Issue #2287, opened 2024-06-03
[bug]: useTransition with a function not working


SageJustus:
### Which react-spring target are you using?

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

latest

### What's Wrong?

The `With a function & deps` approach does not work properly, while the `With a config object` approach works properly.

### To Reproduce

`transitions1` not working

### Expected Behaviour

`transitions1` working

### Link to repo

https://codesandbox.io/p/sandbox/use-spring-1-58c7f3?file=%2Fsrc%2FApp.tsx%3A28%2C10-28%2C22

PR #2289, opened 2024-06-12
fix: useTransition hook works correctly with a function argument. (#2287)

Roman178:
### Why
https://github.com/pmndrs/react-spring/issues/2287.
The useTransition hook doesn't work correctly if we pass a function as an argument.

### What

If we pass a function to the hook, a `ref` variable will be created, an instance of `SpringRef`.
Then, as stated in the comment, if the `ref` exists, the update is postponed until the `ref`'s `start` method is called. By calling the ctrl `update` method, we simply put another item in the queue. Perhaps I didn’t fully understand it, but I didn’t see calling the `start` method on an existing `ref` that flushes the queue.

Calling the `ref`'s `start` method in useIsomorphicLayoutEffect solves the problem.

image

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [ ] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useTransition.tsx
+++ b/packages/core/src/hooks/useTransition.tsx
@@ -422,7 +422,9 @@ export function useTransition(
             }
           }
         }
-      )
+      );
+
+      if (ref) ref.start();
     },
     reset ? void 0 : deps
   )


Roman178:
Hello @joshuaellis 
Just a friendly reminder. I would appreciate it if you would take a moment to review at my PR.

Roman178:
Hello @joshuaellis
Just a friendly reminder. I would appreciate it if you would take a moment to review my PR.

joshuaellis:
Hey sorry for the late reply.

The useTransition hook when provided with a config function isn't designed to play its animations automatically. Instead you receive a second item in the returned array which is the API for you to trigger the animation when desired.

If you want it to automatically play, just pass an object. 

Let me know if I've misunderstood something about the issue you're trying to fix 😊

PR #2293, opened 2024-06-13
support removing attributes from animated elements

LoganDark:
### Why

[`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) does not accept `inert=''`, `inert='false'`, `inert='undefined'` or etc. The attribute ***must*** be entirely removed in order to enable the element again. [`disabled` works the same way](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute).

### What

If a value resolves to `undefined` (not `null` or `false` or otherwise) its attribute will now be entirely removed.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/.changeset/animate-attribute-removal.md
+++ b/.changeset/animate-attribute-removal.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/web': patch
+---
+
+fix: support removing attributes from animated elements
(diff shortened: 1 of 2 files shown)


joshuaellis (changes requested):
Before we can evaluate on this being the right thing to do properly, can we add a couple of test cases to `animated.test.tsx` that shows this working please? :)

LoganDark:
I suppose right now it's impossible for a spring value to even be `null` or `undefined` without an interpolation, because of #2295. I'm off work for the week, so I can look into this more next Monday.

joshuaellis:
@LoganDark can you rebase this please? :)

LoganDark:
oh oops I didn't realize I had another PR open, lmao, one moment

LoganDark:
#2319

Issue #2299, opened 2024-07-01
[bug]: useReducedMotion: error on Safari < 14


BenHodgsonR2R:
### Which react-spring target are you using?

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

9.7.3

### What's Wrong?

The `useReducedMotion` hook calls MediaQueryList.addEventListener(), this causes an error in older Safari browsers- which are yet to have this function available. 

### To Reproduce

- Be on Safari 13.1 or lower
- Visit this url: https://codesandbox.io/s/eager-gould-hm3rtj?file=/src/App.tsx
- See the error

### Expected Behaviour

Doesn't throw an error and also honours the `(prefers-reduced-motion)` media query.

### Link to repo

https://codesandbox.io/s/eager-gould-hm3rtj?file=/src/App.tsx

PR #2300, opened 2024-07-01
fix: add `useReducedMotion` compatibility for Safari 13 and older

BenHodgsonR2R:
### Why

Currently, calling `useReducedMotion` on a Safari browser older than v14 causes an error, crashing the app.
This pull request adds compatibility checks, so that `useReducedMotion` can be called on older devices without producing an error.

Resolves https://github.com/pmndrs/react-spring/issues/2299

### What

Adds truthy checks to `MediaQueryList.addEventListener` before attempting to call the function. This function is undefined on older browsers.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/.changeset/stale-owls-double.md
+++ b/.changeset/stale-owls-double.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/shared': minor
+---
+
+fix: MediaQueryList.addListener safari 13 (and older) compatibility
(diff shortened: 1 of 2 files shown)


joshuaellis:
Can you please add a changeset to the PR? 😊 otherwise looks great to me, thanks for the fix!

BenHodgsonR2R:
@joshuaellis 
Thank you for taking a look at this so quickly!
Just added the changeset

joshuaellis (dismissed):

joshuaellis (approved):

PR #2311, opened 2024-08-12
fix(core): Added threshold field to IntersectionArgs for useInView hook

roman-khasimyanov:
### Why

According to [doc](https://www.react-spring.dev/docs/utilities/use-in-view) `useInView` must accept field `threshold`, but there is type error

### What

fix types

### Checklist

- [x] Documentation updated
- [x] Ready to be merged

### DIFF
--- a/.changeset/good-elephants-matter.md
+++ b/.changeset/good-elephants-matter.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/core': patch
+---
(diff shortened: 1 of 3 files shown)


joshuaellis:
I actually think the docs are wrong, we should instead prefer `amount` to be used instead. I'll amend the docs.

PR #2319, opened 2024-09-16
support removing attributes from animated elements (try 2)

LoganDark:
### Why

[`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert) does not accept `inert=''`, `inert='false'`, `inert='undefined'` or etc. The attribute ***must*** be entirely removed in order to enable the element again. [`disabled` works the same way](https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute).

### What

If a value resolves to `undefined` (not `null` or `false` or otherwise) its attribute will now be entirely removed.

### Checklist

- [x] Ready to be merged


targets/web/src/applyAnimatedValues.ts
```diff
@@ -70,7 +74,12 @@ export function applyAnimatedValues(instance: Instance, props: Lookup) {
...
-    instance.setAttribute(name, values[i])
+    const value = values[i]
+    if (value !== void 0) {
+      instance.setAttribute(name, value)
+    } else {
+      instance.removeAttribute(name)
+    }
   })
 
   if (className !== void 0) {
```
joshuaellis, line 85: should we apply the same logic of removing to all the props here? Just wondering the rationale behind the ones you've changed at the moment.

PR #2357, opened 2025-02-09
Clarify what `immediate` does

jorisw:
### Why

The naming of `immediate` _suggests_ target styles will be applied instead of running the animation, but this isn't explicitly stated. This makes me unsure of whether I should be using `immediate` when wanting to skip motion.

### What

Clarifies what `immediate` does to 'prevent' animation

### Checklist

- [x] Documentation updated
- [ ] Demo added
- [x] Ready to be merged


joshuaellis (approved):

joshuaellis:
Can you run `prettier` please?

PR #2363, opened 2025-03-20
Feat: React 19 Upgrade

kierancap:
### Why

Closing https://github.com/pmndrs/react-spring/issues/2341

### What

Main changes:

- Removed all usage of `forwardRef` - As per the React docs, will be fully deprecated in future.
- Replaced all `ref` fields with the valid React Types - Any polymorphic components were set as `any` for the `RefObject` generic.
- Added valid initial values for all usages of `useRef` which were missing inits.
- Refactored the `SpringContext` to resemble a more sensible Context structure

### Checklist

- [x] Removed forwardRef
- [x] Added correct `ref` fields
- [x] Added initial values for `useRef`
- [x] Refactored SpringContext

### Additional Comments

This was based on this PR: https://github.com/pmndrs/react-spring/pull/2349

As the PR and issue have been open for a good while now, it's obvious that no further development was going to happen - I'm currently blocked by this library, so I thought it'd try and help with moving this forward.

### PR Diff

I'm unsure as to why the diff is heavily focused on formatting changes - I was sure to run prettier via the NPM script before submitting this PR. Happy to outline all the direct code changes if necessary.

PR #2368, opened 2025-04-18
feat!: react 19 support

dagatsoin:
This first is up to date from the below discussion so you don't waste time to read all the thread**

## --> CURRENT RELEASE: **10.0.0-beta.0** <--

TLDR;
### Todo

- [x] Lint OK
- [x] Build OK
- [x] Tests OK
- [x] Documentation updated
- [x] Web target update
- [x] review: peer dependencies ([need reviewer!](https://github.com/pmndrs/react-spring/pull/2368#discussion_r2056720676))
- [x] review: tsconfig update for unsupported compiler option ([discussion](https://github.com/pmndrs/react-spring/pull/2368#discussion_r2060480490))
- [x] review: demo/target dependencies ([need reviewer!](https://github.com/pmndrs/react-spring/pull/2368/files#r2060466971))
- [x] review: React Native build ([self reviewed](https://github.com/pmndrs/react-spring/pull/2368#discussion_r2056804269))
- [x] v10 beta release
- [ ] ~~integration test on a ZDog real project~~ [react-zdog is not compatible yet with React 19](https://github.com/pmndrs/react-spring/pull/2368#issuecomment-2853642931)
- [x] integration on a Konva real project [test by lib author](https://github.com/pmndrs/react-spring/pull/2368#issuecomment-2839533619)
- [x] integration on a React Native real project ([exemple here](https://dagatsoin.github.io/sproutch/iframe.html?globals=&id=advanced-animated-transition--basic&viewMode=story))

### Why

Support for React 19.

This PR is the last attempt in date to support React 19. It follows the tremendous effort which have been poured into #2363—as a great example of community effort. Kudo [kierancap](https://github.com/kierancap) for the work on the Context.

### Notable changes

BREAKING CHANGE:
- update(context)!: <SpringContext ...> is not usable anymore. You need to update your project to use <SpringContextProvider ...>

OTHER CHANGES:
- chore(native): remove ES export and use 'exports' field in package.json to comply to Metro 0.82
- chore(peer): accepts react/react-dom from 16.8 to <20
- chore(peer): update konda and zdog to their react 19 compatible version
- chore(deps): set demo/ci/docs to react/react-dom 19
- chore(devDeps): set to react/react-dom 19
- refactor: impact due to the upgrade to React 19 (linting, typing)


kierancap:
Hey! Yeah the linting/formatting parts of my PR caused me a bit of a headache too... I did run the formatting script before committing anything, which I would have thought would solve the problem! I guess not!

I would actually say that the underlying problem may be either my individual setup, or the global setup for linting/formatting in the package as a whole. (From personal experience I know how difficult it is to manage it all properly in a massive monorepo).

Do you think it'd be easier to wait for my PR to be merged - then applying the formatting fixes? (I'd actually be interested to see if you got any different results from running the `pretter:check/write` scripts in my branch)

Realistically if the issue is stemming from a formatting mismatch from either my machine or the monorepo as a whole - Fixing that may help to resolve the problem?

Obviously i'm happy to pull out the core parts of my PR into this PR if that's truly necessary

dagatsoin:
@kierancap thanks for you answer.

After looking back at your PR, it seems that the formatting is a smaller problem since your latest commits. There are way fewer space changes. So this is not a problem anymore.

I rebased your work on my PR so here is a change log. So you have the choice to work on this one or yours.

## Fixed:

### Use of "any"

We were losing a lot of benefit from TypeScript static typing.

This was caused by the partial changes to the peer and dev dependencies. There are a lot of packages to update including React, React-dom, drei, and three-fiber which prevents the need to "any" all the "Tag" related types.

### Using null over undefined

Some "useRef" are initialized with "null" while they were initialized with implicit undefined.

before: `useRef()`
now:  `useRef(null)`

which is different from `useRef(undefined)` undefined !== null

## To fix:

## Potential breaking change on the context.

I rebased the work on jest setup and SprintContext so all the tests pass now.

But there is a catch.

Those changes were made:

(Note : I use `ISpringContext` for the interface to diff from the `SpringContext` object.)

The current version exports a context object with the following signature :
`type  SprintContext = React.Context<ISpringContext> & React.FC<PropsWithChildren<ISpringContext>>`

But now it splits the exported context in two parts:
`type SpringContextProvider = React.Provider<ISpringContext>`
and
`type SpringContext = React.Context<ISpringContext>`

Given that the context provider was exported at the package level, it will break any project that uses `SpringContext`.

Ex: If my project uses the old version, switching to this new version will break this code:

```tsx
import { SpringContet } from 'react-spring';

export myComp = function({children, ...props}) {
  return <SpringContext {...props}>
      {children}
    </SpringContext>
}
```

So, either we bump to a major version due to the breaking change, or we stay at patch/minor level and find a way to keep the same package interface.

kierancap:
> @kierancap thanks for you answer.
> 
> 
> 
> After looking back at your PR, it seems that the formatting is a smaller problem since your latest commits. There are way fewer space changes. So this is not a problem anymore.
> 
> 
> 
> I rebased your work on my PR so here is a change log. So you have the choice to work on this one or yours.
> 
> 
> 
> ## Fixed:
> 
> 
> 
> ### Use of "any"
> 
> 
> 
> We were losing a lot of benefit from TypeScript static typing.
> 
> 
> 
> This was caused by the partial changes to the peer and dev dependencies. There are a lot of packages to update including React, React-dom, drei, and three-fiber which prevents the need to "any" all the "Tag" related types.
> 
> 
> 
> ### Using null over undefined
> 
> 
> 
> Some "useRef" are initialized with "null" while they were initialized with implicit undefined.
> 
> 
> 
> before: `useRef()`
> 
> now:  `useRef(null)`
> 
> 
> 
> which is different from `useRef(undefined)` undefined !== null
> 
> 
> 
> ## To fix:
> 
> 
> 
> ## Potential breaking change on the context.
> 
> 
> 
> I rebased the work on jest setup and SprintContext so all the tests pass now.
> 
> 
> 
> But there is a catch.
> 
> 
> 
> Those changes were made:
> 
> 
> 
> (Note : I use `ISpringContext` for the interface to diff from the `SpringContext` object.)
> 
> 
> 
> The current version exports a context object with the following signature :
> 
> `type  SprintContext = React.Context<ISpringContext> & React.FC<PropsWithChildren<ISpringContext>>`
> 
> 
> 
> But now it splits the exported context in two parts:
> 
> `type SpringContextProvider = React.Provider<ISpringContext>`
> 
> and
> 
> `type SpringContext = React.Context<ISpringContext>`
> 
> 
> 
> Given that the context provider was exported at the package level, it will break any project that uses `SpringContext`.
> 
> 
> 
> Ex: If my project uses the old version, switching to this new version will break this code:
> 
> 
> 
> ```tsx
> 
> import { SpringContet } from 'react-spring';
> 
> 
> 
> export myComp = function({children, ...props}) {
> 
>   return <SpringContext {...props}>
> 
>       {children}
> 
>     </SpringContext>
> 
> }
> 
> ```
> 
> 
> 
> So, either we bump to a major version due to the breaking change, or we stay at patch/minor level and find a way to keep the same package interface.
> 
> 
> 
> 
> 
> 
> 
> 

Hey!

So...

## Any

The any was due to polymorphic components (if i'm remembering right - really annoying). If you were able to tidy those up that's great!

## Refs

Completely agreed on the approach to remove `undefined` from useRef.

## Context

I'm nearly certain that with the newest types derived from React, that merging a Provider AND Context initiator into the same object will be nigh on impossible. That's why that old setup was so un-idiomatic and seemed to chip away at the underlying context logic to allow for it to happen.

I would definitely consider a Major bump here - Josh also agreed on the context change in my initial PR. 

The Provider paradigm for Context is quite popular these days, and is way simpler on the DX to use. Obviously it does give us a breaking change, but for overall maintainability I think that's a valued change

dagatsoin:
@kierancap about the breaking change: indeed, I found the message where @joshuaellis agrees. 

How do you want to proceed? Either you cherry pick from mine and I close this one or we continue on this one.

It remains:

- [x] document the breaking change
- [x] test with react-native

dagatsoin:
Assuming the breaking change, the minimum requirements is now React 19
- updated react-native to 0.78
- updated react-konda to 19

dagatsoin:
added a breaking change section in doc

dagatsoin:
fix types error in demo due to mutliple react version:
- updated styles-component
- force to use newer version of react-transition-group (dependency of react-select)

kierancap:
> @kierancap about the breaking change: indeed, I found the message where @joshuaellis agrees. 
> 
> 
> 
> How do you want to proceed? Either you cherry pick from mine and I close this one or we continue on this one.
> 
> 
> 
> It remains:
> 
> 
> 
> - [x] document the breaking change
> 
> - [ ] test with react-native 

I personally feel like you're making really good progress on this - So if you want to pick up where I left off (full transparency, I won't have much time to keep up with this for the next few weeks). I'm happy for you to continue your work on this one.

I can alert on my PR that we move to this one and I can close my PR. 

Would that suit you? If you're happy to adopt my context change (thats the only real major change to the upgrade from my PR - ignoring the test fixes and the useRef init values)

Thank you mate! 

(The alternative is that we keep my branch active and you can submit PRs to that branch directly if you wish)

dagatsoin:
@kierancap ok, I can continue from here. I will check your last commit but I think I am up to date.

packages/core/package.json
```diff
@@ -57,6 +57,6 @@
     "@react-spring/types": "~9.7.5"
   },
   "peerDependencies": {
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0"
+    "react": "^19"
```
joshuaellis: Why isn't this backwards compatible? I think a lot of other libs have been able to keep that 🤔

dagatsoin: @joshuaellis  Yes good catch. I was confused about the breaking change brought by the new SpringContext api.
To confirm, useContext is well supported in React 16.8?

If so, I propose to set 
"react": ">=16.8 <20"

joshuaellis: Yes it is, imo go with the same format we had originally so 
```
    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
```

dagatsoin:
## Crash on my real life tests on React Native:

`import { animated } from '@react-spring/native';`

- passes with React Native 0.76
- breaks with React Native > 0.78

## Cause:

The package @react-spring/native is not resolved with the right path by Metro. Indeed, the ESM package is used which is not compatible with RN (RN uses only CommonJS).

It is due to the [Metro breaking change since v0.82](https://metrobundler.dev/docs/package-exports#recommended-introducing-exports-is-a-breaking-change), about the "exports" support in package.json.

Metro now uses the exports declaration prior to other entry points. So:
- on Metro < 0.82 the resolver uses the "main" entry point in the package.json which resolves to the CJS module
- on Metro >= 0.82 the resolver uses the "exports" entry point in the package.json which resolves to the ESM module

## Impact:
- All RN projects using >= v0.78 will break.

## Proposal

To handle that we need to find two entry points in package.json.

**EDIT: this proposal has been successfully tested in a client project.**

@joshuaellis I have a proposal on this: as far as I know, Metro only uses CommonJS as module system, so:
- we remove the compilation in ES module in react-spring/native
- we set the following in targets/native/package.json

```json 
"main": "./dist/cjs/index.js",
"types": "./dist/cjs/react-spring_native.development.d.ts",
"exports": {
    "./package.json": "./package.json",
    ".": {
      "types": "./dist/cjs/react-spring_native.development.d.ts",
      "default": "./dist/cjs/index.js"
    }
  },
```

packages/core/package.json
dagatsoin: for checking and resolve: @joshuaellis

targets/native/tsup.config.ts
```diff
@@ -6,6 +6,7 @@ export default defineConfig(opts => {
     {
       entry: 'src/index.ts',
       name: 'react-spring_native',
+      buildFilter: ['development', 'production.min']
```
dagatsoin: Added a filter to build only what needed (use case: React Native just needs CJS format)

tsup.config.base.ts
```diff
@@ -91,13 +91,16 @@ if (process.env.NODE_ENV === 'production') {
...
 }
 
 export const defaultConfig = (
-  { name: prefix, entry }: ConfigOptions,
+  { name: prefix, entry, buildFilter }: ConfigOptions,
   options: Options
 ): Options[] => {
-  const artifactOptions: Options[] = buildTargets.map(
+  const artifactOptions: Options[] = buildTargets
+    .filter(target => !buildFilter || buildFilter.includes(target.name))
```
dagatsoin, line 102: @joshuaellis about removing the ESM build for React Native.

 I use the filter here but I have an issue. I am new to turbo, so I surely miss something.

Here is a view of my logs during the build. In pink, the log for the native target show that the build are well filtered.
image

But the ESM flavor files still lands in the dist folder.

image

Have you any idea why?

dagatsoin:
gentle ping @joshuaellis

tsup.config.base.ts
joshuaellis: Sorry why do we want to remove ESM from react-native?

Is that what other libraries do? That feels weird.

targets/native/tsup.config.ts
dagatsoin: @joshuaellis I explain it here https://github.com/pmndrs/react-spring/pull/2368#issuecomment-2824536087

tsup.config.base.ts
dagatsoin: I explain it here https://github.com/pmndrs/react-spring/pull/2368#issuecomment-2824536087

dagatsoin: @joshuaellis I explain this here https://github.com/pmndrs/react-spring/pull/2368#issuecomment-2824536087

tsconfig.json
```diff
@@ -14,14 +14,14 @@
   ],
   "compilerOptions": {
     "target": "ES2020",
-    "module": "ESNext",
+    "module": "NodeNext",
```
dagatsoin, line 17: As we need to get rid of ESM in the native target, we need to handle both ESM and CJS output format. Which is supported by NodeNext.

Otherwise, @react-sping/native definition will not be found at code time and crashes on tests and lint.

tsup.config.base.ts
dagatsoin: Solved [here](https://github.com/pmndrs/react-spring/pull/2368/files#diff-b53466a39825cc473ab3e1acaea45dddf86aa08fe69a3ab8bd38c3193dfcc361R150) @joshuaellis 
The esOptions being present, it seems to automatically trigger the ES build regardless of the module resolution context. So I put a condition inside.

Test still passes, no regression detected.

targets/three/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "peerDependencies": {
     "@react-three/fiber": ">=6.0",
```
dagatsoin, line 53: Do we need to bump here ?

.github/publish-ci/cra5/package.json
```diff
@@ -4,8 +4,8 @@
   "private": true,
   "dependencies": {
     "@react-spring/web": "*",
-    "react": "18.3.1",
-    "react-dom": "18.3.1",
+    "react": "^19.1.0",
```
dagatsoin, line 7: NEED REVIEWER HERE

Changes:
- [x] All dev dependencies bumped to React 19
- [x] All demo/ci dependencies bumped to React 19
- [x] styledComponent updated to support React 19
- [x] update demo: @react-three/drei update to support React 19
- [x] update demo: @react-three/fiber update to support React 19
- [x] update demo: @react-select update to support React 19

demo/src/sandboxes/animating-auto/tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "noImplicitThis": true,
     "noImplicitAny": true,
     "strictNullChecks": true,
-    "suppressImplicitAnyIndexErrors": true,
```
dagatsoin, line 16: NEED REVIEWER HERE

- [x] removed suppressImplicitAnyIndexErrors from ts to get rid of this error in all tsconfig.json

image

joshuaellis: cool 👍🏼

joshuaellis:
`10.0.0-beta.0` should be released for trying now.

demo/src/sandboxes/animating-auto/tsconfig.json
dagatsoin: Assuming it is resolved. Don't hesitate to reopen if it is not.

lavrton:
Tested the `10.0.0-beta.0` version with Konva integration and the latest react/react-dom/react-konva versions. The minimal example is working just fine: https://codesandbox.io/p/sandbox/compassionate-hodgkin-7h4wvl?file=%2Fsrc%2Findex.js

dagatsoin:
About React Native, I confirm it works very well. I work with for a week in fast iteration for my Sproutch library and have no issue. I allow myself to validate the build point.

tsup.config.base.ts
dagatsoin: I have already build a few pre release versions of my Sproutch UI library including online demos:
- RN example: exp+://expo-development-client/?url=https://u.expo.dev/29b4794d-fe0b-48d8-8705-26b7751d96cd/group/c3510727-a9b2-4c55-a666-8ff753329ddf
- [Web example](https://dagatsoin.github.io/sproutch/?path=/docs/introduction--docs)
My stack: expo + storybook + RN web.

dagatsoin:
@joshuaellis it appears that zdog support does not work.
react-zdog [does not support yet React 19](https://github.com/pmndrs/react-zdog/blob/87c99ae1650265e02590b0f682302ba03aac6409/package.json#L31) so this error pops up

image

I added a warning in the @react-spring/zdog readme.

joshuaellis:
> @joshuaellis it appears that zdog support does not work. react-zdog [does not support yet React 19](https://github.com/pmndrs/react-zdog/blob/87c99ae1650265e02590b0f682302ba03aac6409/package.json#L31) so this error pops up
> 
> image
> I added a warning in the @react-spring/zdog readme.

Makes sense!

VincentHui:
At least for web I've used it with react 19 and chakra ui and have had no issues so far

joshuaellis (dismissed):

joshuaellis (approved):

Issue #2371, opened 2025-05-14
[bug]: Crash to white screen


leo-petrucci:
### Which react-spring target are you using?

- [ ] `@react-spring/web`
- [ ] `@react-spring/three`
- [x] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

10.0.0

### What's Wrong?

I'm trying to use `react-spring` in an Expo React Native app however the app crashes to white screen as soon as I open it.

### To Reproduce

1. I installed `@react-spring/native`
2. I imported `import { useSpring, animated } from "@react-spring/native";`
3. Add `useSpring` to a component
4. App crashes with white screen.

### Expected Behaviour

App shouldn't crash.

### Link to repo

https://github.com/leo-petrucci/react-spring-native-bug/tree/master

leo-petrucci:
I just noticed `10.0.0` got released. I updated and now I'm only getting a white screen :/

dagatsoin:
@leo-petrucci I can reproduce it indeed :(
It affect also the web version. I have no work around for now. I will try to take a look this evening.

leo-petrucci:
@dagatsoin Thank you! I think the initial issue was caused by me trying to use it with React 19. I'll update the OP/Title

joshuaellis:
resolved by #2372

dagatsoin:
@joshuaellis I don't think the last fix solves this, it seems to be an other root cause.

dagatsoin:
On my test, the loop property blocks the process, so the blank page. Happens on web and native.

dagatsoin:
@leo-petrucci there is a work around.
It seems to work if you use the function props. In your case:

```ts
useSpring(() => ({
    from: { rotate: "0deg" },
    to: [
      { rotate: "0deg" },
      { rotate: "-5deg" },
      { rotate: "5deg" },
      { rotate: "0deg" },
    ],
    config: { duration: 100 },
    loop: true,
}), []);
```

leo-petrucci:
This is working for me without the workaround mentioned by @dagatsoin 👀

Thank you for fixing this!

dagatsoin:
@leo-petrucci ok that's weird but happy to read!
Sorry @joshuaellis you can re-close it. The bug persists for me, but the root cause may be elsewhere 🛸

hrishikesh2810:
Hi maintainers 👋
I’d like to work on this issue.
[...]

PR #2372, opened 2025-05-15
fix: spring does not start

dagatsoin:
### Why

This PR fixes a bug where useSpring lead to unmount component (eg. blank page)

Related: #2371

### What

The root cause was an instruction which removes the current update during the useIsomorphicLayoutEffect callback.

@kierancap I would like your opinion on this if you have a minute. I don't know exactly the impact of this.

### Checklist

- [x] Demo added: check before/after the demo css-variables
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useSprings.test.tsx
+++ b/packages/core/src/hooks/useSprings.test.tsx
@@ -29,7 +29,7 @@ describe('useSprings', () => {
     it('should reach final value in strict mode', async () => {
       update(1, () => ({
         from: { x: 0 },
-        to: { x: 1 }
+        to: { x: 1 },
       }))
       expect(mapSprings(s => s.goal)).toEqual([{ x: 1 }])
     })
--- a/packages/core/src/hooks/useSprings.test.tsx
+++ b/packages/core/src/hooks/useSprings.test.tsx
@@ -26,6 +26,14 @@ describe('useSprings', () => {
   }, isStrictMode)
 
   describe('when only a props function is passed', () => {
+    it('should reach final value in strict mode', async () => {
+      update(1, () => ({
+        from: { x: 0 },
+        to: { x: 1 }
+      }))
+      expect(mapSprings(s => s.goal)).toEqual([{ x: 1 }])
+    })
+
     it('calls the props function once per new spring', () => {
       const getProps = jest.fn((i: number) => ({ x: i * 100 }))
 
--- a/packages/core/src/hooks/useSprings.ts
+++ b/packages/core/src/hooks/useSprings.ts
@@ -218,7 +218,6 @@ export function useSprings(
           ctrl.start(update)
         }
       }
-      updates.current[i] = null
     })
   })
 
--- a/packages/core/src/hooks/useSprings.ts
+++ b/packages/core/src/hooks/useSprings.ts
@@ -127,6 +127,10 @@ export function useSprings(
     []
   )
 
+  // useRef is needed to get the same references accross renders.
+  // Note that controllers are a shallow copy of the final array.
+  // So any array manipulation won't impact the final array until
+  // the commit phase, in the useIsomorphicLayoutEffect callback
   const ctrls = useRef([...state.ctrls])
   const updates = useRef<any[]>([])
 
@@ -180,6 +184,10 @@ export function useSprings(
   const prevContext = usePrev(context)
   const hasContext = context !== prevContext && hasProps(context)
 
+  // This is the commit phase where the new transition will begin.
+  // - updated animation values will be passed to the controlers.
+  // - state will be updated
+  // - springs will start or the new update will be queued if the spring is not started yet.
   useIsomorphicLayoutEffect(() => {
     layoutId.current++
 
--- a/packages/core/src/hooks/useSprings.ts
+++ b/packages/core/src/hooks/useSprings.ts
@@ -129,7 +129,6 @@ export function useSprings(
 
   const ctrls = useRef([...state.ctrls])
   const updates = useRef<any[]>([])
-  updates.current ??= []
 
   // Cache old controllers to dispose in the commit phase.
   const prevLength = usePrev(length) || 0


joshuaellis:
Do we have a reproduction of this in a sandbox? if i look at the examples in the docs, they all appear to work 🤔

dagatsoin:
@joshuaellis yes,
- for the version with a bug : [css-variables demo](https://codesandbox.io/p/sandbox/8x50e?file=%2Fpackage.json%3A15%2C27)
- for the version without a bug : compile the demo of my branch and launch the css-variables demo

kierancap:
I'll take a look at this properly in a few hours. That line in particular was taken from the stale PR we had for React 19 before mine. I assumed it was fixing a cleanup of the ref. Would be good to test properly without it. 

Whatever the cause of this bug is - Seeing as the CI tests passed, we should try and enhance the coverage to try and replicate this via CI. 

If you were able to recreate the bug consistently is there a particular case where it fails? 

As Josh mentioned, the local tests and the deployed docs worked! Odd! 

Thoughts?

kierancap:
I've just had a proper look through - I think we can remove this safely - I don't think the cleanup is necessary as the next runthrough of updates will overwrite any previously parsed updates. So in terms of the logic for removing it, all good on my end.

One thing I would like to suggest though is to catch the crashing behaviour in a test (like i mentioned before). I guess this is dependant on what causes the crashing, i.e what error it is. What are you guys' thoughts on that? Worth doing?

dagatsoin:
@kierancap agree to add a test on this, but I would first document the internal a bit. It is not easy to grasp for everyone.
The test will be more relevant.

So, what this business code actually do?

PS: the current website is not up to date so it run with the v9 
You have a link upper to see a repro case.

kierancap:
Yeah agreed about documenting the internals. Based on my understanding the ref is just handling queuing updates. I think the cleanup line may have come from the thought of needing the clean up an update post process, but based on my glance yesterday it seems like thats unnecessary (I'm also guessing that the ref is used asynchronously in other parts too - hence the crashing)

As to the actual design of the code, I'm not totally sure - It's not well documented so I've just tried to peace together what I can based on the code itself.

The last updates to that system pre our PR was 5-6 years ago, so I doubt we can get any assistance on that either. 

If @joshuaellis has any understanding about the spring update system, would be great to enlighten myself and we can write some docs outlining it internally

joshuaellis:
> If @joshuaellis has any understanding about the spring update system, would be great to enlighten myself and we can write some docs outlining it internally

It was written by a previous maintainer, i normally have to read it a couple of times before i really understand it. 

I think the point about it not being caught by a test is important, i would revert the change temporarily and get a test to fail then go from there. It might be that this is caught in a more integration test where you render a component instead of just the hook perhaps?

kierancap:
Ah that's fair enough! 

Would absolutely be worth testing this from multiple angles, seeing as it escaped us the first time! 

I'd be interested to see if the bug can be triggered via the hook alone - or if it's dependant on a full React environment (UI rendering etc) to appear

dagatsoin:
some more in depth explanations @joshuaellis @kierancap

TL;DR
On the second render, declareUpdates is not called because the dependencies of the memo hooks do not change.
As a result, there is no controller, and the animation does not start.

======

After diving back into the code, it's clear that the cleanup line makes no sense. I'm pretty sure it was just a WIP line meant for testing, which accidentally ended up in our PR.

## What I learned
### Bug reproduction environment
With v10, it does not appear in production mode, regardless of the following conditions:
- With or without useRef
- With or without the cleanup line
- With React 18 or 19

### Usage: the updates and controllers arrays

These are internal state structures:

`updates` stores the successive states the animation should reach. (e.g., when you click repeatedly for an element to follow your mouse, each click adds a new update with a new destination)

`ctrls` are the controller instances handling each step of the animation

### Logic: starting the animation

For a spring, starting an animation means sending the next state to reach.
In our code, this happens at [line 209](https://github.com/pmndrs/react-spring/pull/2372/files#diff-e1288b0efcce202f0575295aec86c6f33c5d81adac9751131dc5d2b08e40e73aR209).
If there is a new update, we start the animation.

### Usage:  useRef

In Strict Mode, React triggers two renders.

When there's only one render, everything works fine
- `updates` and `ctrls` are populated during the useMemo hooks
-  [line 209](https://github.com/pmndrs/react-spring/pull/2372/files#diff-e1288b0efcce202f0575295aec86c6f33c5d81adac9751131dc5d2b08e40e73aR209) condition is true
- the spring starts.

But on the second render, since the useMemo dependencies haven't changed, the useMemo hooks are not triggered, resulting in empty updates and ctrls arrays.

Using useRef (and also useMemo to store ctrls) allows us to keep references to previously populated arrays.

## The bug

The bug originates from the cleanup line where we remove an update.
In Strict Mode, this means:
- First render
- `useMemo` is triggered, and updates/ctrls are populated
- The queue starts processing inside the useIsomorphicLayoutEffect callback:
  - The condition at line 209 is true, so the spring starts
  - The same update is removed from the array
- Second render
- useMemo is triggered, but updates/ctrls remain unchanged (💀 )
- The queue starts processing again in the useIsomorphicLayoutEffect callback:
  - The condition at line 209 is false because the first update was already removed during the previous render and was not be restored during the useMemo hooks

🐛🐛🐛 The spring won't start 🐛 🐛🐛

## Solution

Removing the faulty cleanup line.

kierancap:
> some more in depth explanations @joshuaellis @kierancap
> 
> 
> 
> TL;DR
> 
> On the second render, declareUpdates is not called because the dependencies of the memo hooks do not change.
> 
> As a result, there is no controller, and the animation does not start.
> 
> 
> 
> ======
> 
> 
> 
> After diving back into the code, it's clear that the cleanup line makes no sense. I'm pretty sure it was just a WIP line meant for testing, which accidentally ended up in our PR.
> 
> 
> 
> ## What I learned
> 
> ### Bug reproduction environment
> 
> With v10, it does not appear in production mode, regardless of the following conditions:
> 
> - With or without useRef
> 
> - With or without the cleanup line
> 
> - With React 18 or 19
> 
> 
> 
> ### Usage: the updates and controllers arrays
> 
> 
> 
> These are internal state structures:
> 
> 
> 
> `updates` stores the successive states the animation should reach. (e.g., when you click repeatedly for an element to follow your mouse, each click adds a new update with a new destination)
> 
> 
> 
> `ctrls` are the controller instances handling each step of the animation
> 
> 
> 
> ### Logic: starting the animation
> 
> 
> 
> For a spring, starting an animation means sending the next state to reach.
> 
> In our code, this happens at [line 209](https://github.com/pmndrs/react-spring/pull/2372/files#diff-e1288b0efcce202f0575295aec86c6f33c5d81adac9751131dc5d2b08e40e73aR209).
> 
> If there is a new update, we start the animation.
> 
> 
> 
> ### Usage:  useRef
> 
> 
> 
> In Strict Mode, React triggers two renders.
> 
> 
> 
> When there's only one render, everything works fine
> 
> - `updates` and `ctrls` are populated during the useMemo hooks
> 
> -  [line 209](https://github.com/pmndrs/react-spring/pull/2372/files#diff-e1288b0efcce202f0575295aec86c6f33c5d81adac9751131dc5d2b08e40e73aR209) condition is true
> 
> - the spring starts.
> 
> 
> 
> But on the second render, since the useMemo dependencies haven't changed, the useMemo hooks are not triggered, resulting in empty updates and ctrls arrays.
> 
> 
> 
> Using useRef (and also useMemo to store ctrls) allows us to keep references to previously populated arrays.
> 
> 
> 
> ## The bug
> 
> 
> 
> The bug originates from the cleanup line where we remove an update.
> 
> In Strict Mode, this means:
> 
> - First render
> 
> - `useMemo` is triggered, and updates/ctrls are populated
> 
> - The queue starts processing inside the useIsomorphicLayoutEffect callback:
> 
>   - The condition at line 209 is true, so the spring starts
> 
>   - The same update is removed from the array
> 
> - Second render
> 
> - useMemo is triggered, but updates/ctrls remain unchanged (💀 )
> 
> - The queue starts processing again in the useIsomorphicLayoutEffect callback:
> 
>   - The condition at line 209 is false because the first update was already removed during the previous render and was not be restored during the useMemo hooks
> 
> 
> 
> 🐛🐛🐛 The spring won't start 🐛 🐛🐛
> 
> 
> 
> ## Solution
> 
> 
> 
> Removing the faulty cleanup line.

Incredible work diagnosing this!! 

It's so rare to see actual occasions where Strict Mode causes bugs like this, but that's what it's there for!

You are probably absolutely correct that the line was probably there for debugging rather than an actual addition to the code. If the bug is purely caused by the ref being removed and Strict Mode causing the oddness, there's not much value in testing that scenario. (I saw a commit about a test, if that's already in place, feel free to keep it! I just know from personal experience that testing Strict Mode oddness is a challenge)

I'd be happy for this PR to be moved forward in that case. If @joshuaellis doesn't have any further comments of course :) 

Great work again mate!

dagatsoin:
thx @kierancap!
In fact, the test is really simple and the case was not covered.
You can test it if you want in the current branch by re copying the delete cleanup line to fail it.

joshuaellis (approved):
Amazing work 👏🏼

PR #2373, opened 2025-05-21
fix: remove accessing elem.ref in renderTransitions

robertjarske:
### Why

Accessing element.ref is deprecated. Refs are now regular props and should be accessed using element.props.ref. Accessing element.ref produces a web console error.

https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref

### What

Removed accessing elem.ref in useTransition.tsx. Props are already spreaded.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useTransition.tsx
+++ b/packages/core/src/hooks/useTransition.tsx
@@ -433,18 +433,18 @@ export function useTransition(
         const { springs } = changes.get(t) || t.ctrl
         const elem: any = render({ ...springs }, t.item, t, i)
 
-        if (!elem || !elem.type) return elem
-
         const key = is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id
         const isLegacyReact = React.version < '19.0.0'
 
-        return (
-          <elem.type
-            {...elem.props}
-            key={key}
-            {...(isLegacyReact && { ref: elem.ref })}
-          />
-        )
+        const props = {
+          ...elem.props,
+        }
+
+        if (isLegacyReact) {
+          props.ref = elem.ref
+        }
+
+        return elem && elem.type ? <elem.type key={key} {...props} /> : elem
       })}
     </>
   )
--- a/packages/core/src/hooks/useTransition.tsx
+++ b/packages/core/src/hooks/useTransition.tsx
@@ -432,14 +432,18 @@ export function useTransition(
       {transitions.map((t, i) => {
         const { springs } = changes.get(t) || t.ctrl
         const elem: any = render({ ...springs }, t.item, t, i)
-        return elem && elem.type ? (
+
+        if (!elem || !elem.type) return elem
+
+        const key = is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id
+        const isLegacyReact = React.version < '19.0.0'
+
+        return (
           <elem.type
             {...elem.props}
-            key={is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id}
-            ref={React.version < '19.0.0' ? elem.ref : elem.props.ref}
+            key={key}
+            {...(isLegacyReact && { ref: elem.ref })}
           />
-        ) : (
-          elem
         )
       })}
     </>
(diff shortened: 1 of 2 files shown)


dagatsoin:
Thank you @robertjarske

I can reproduce the error message.

Repro with React 19 ✅ 
image

Note that it is not really an error. React is a bit dramatic about it, it is just [a deprecation](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#new-deprecations).

But after further testing this solution is not retro compatible with React 18. And the v10 still supports it.

Removing the mechanism in useTransition, as you can see below, breaks the React 18 compatibility, the log show an undefined ref.

QA under react 18 ❌  
image

QA under react 19 ✅ 
image

If you really want to get rid of the warning, maybe a solution could be to have a fallback to `element.props.ref` somehow.

robertjarske:
Thank you for checking @dagatsoin 

We could get the react version and have a check that solves this issue, what do you think about something like this?
```javascript
//tsconfig.json
{
...
  "compilerOptions": {
    ...
    "resolveJsonModule": true
  }
}

//useTransition.tsx
import { version as reactVersion } from 'react/package.json'
...

const renderTransitions: TransitionFn = render => (
    <>
      {transitions.map((t, i) => {
        const { springs } = changes.get(t) || t.ctrl
        const elem: any = render({ ...springs }, t.item, t, i)
        return elem && elem.type ? (
          <elem.type
            {...elem.props}
            key={is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id}
            ref={reactVersion < "19.0.0" ? elem.ref : elem.props.ref}
          />
        ) : (
          elem
        )
      })}
    </>
  )
  ```

**Comment by dagatsoin** [comment]
mmh I am not use to this. I wonder if it works when you use the module file directly, like with https://unpkg.com/

**Comment by robertjarske** [comment]
> mmh I am not use to this. I wonder if it works when you use the module file directly, like with https://unpkg.com/

both getting the version from package.json and using React.version seems to be working from what I can see

Using this 
```
diff --git a/dist/react-spring_core.modern.mjs b/dist/react-spring_core.modern.mjs
index 631624fb630980556d47f13cb52a16fa09e22950..017bed34b0b0db1def4be96f7e654380e1bc4876 100644
--- a/dist/react-spring_core.modern.mjs
+++ b/dist/react-spring_core.modern.mjs
@@ -1863,6 +1863,7 @@ function useTrail(length, propsArg, deps) {
 
 // src/hooks/useTransition.tsx
 import * as React2 from "react";
+import {version as reactVersion} from "react/package.json"
 import { useContext as useContext3, useRef as useRef2, useMemo as useMemo3 } from "react";
 import {
   is as is11,
@@ -1874,6 +1875,8 @@ import {
   useIsomorphicLayoutEffect as useIsomorphicLayoutEffect4
 } from "@react-spring/shared";
 function useTransition(data, props, deps) {
+console.log("React.version: ", React2.version)
+console.log("Version from package.json: ", reactVersion)
   const propsFn = is11.fun(props) && props;
   const {
     reset,
```

I get this when running 18.3.1 for both react and react-dom
image

And this when running 19.1.0 for both react and react-dom
image


Neither @types/react or @types/react-dom have been updated here and remain at their previous versions

dagatsoin:
Does React.version works also when the file comes from unpkg ? If so, I would go for it. @joshuaellis what do you think about this ?

robertjarske:
> Does React.version works also when the file comes from unpkg ? If so, I would go for it. @joshuaellis what do you think about this ?

Seems to be working as intended with unpkg as well
https://codesandbox.io/p/sandbox/3hzwf6

image

dagatsoin:
let @joshuaellis say the final work now

joshuaellis (dismissed):
I think this is probably fine, it doesn't seem like there's advise on doing it a different way for library maintainers at the moment. We can always tweak in the future.

robertjarske:
the tests failed with a "Cannot read properties of null (reading 'props')" 
should be fixed now @joshuaellis

joshuaellis (approved):

alexiuscrow:
wanting this myself!

colbywhite:
Love that this was merged. Has it been released yet? `10.0.1` doesn't seem to include the last couple of PRs. Any plan to release that soon @joshuaellis?

joshuaellis:
Thanks for the reminder, this is released in 10.0.2

onordgren:
Hello! I started getting this error after upgrading to 10.0.2, using the useTransition hook.
```TypeError: Cannot read properties of null (reading 'ref')```

PR #2394, opened 2025-09-08
Fix issue with Jest and legacy React apps using useTransition

dimitris-tsetsonis:
### Why

Using latest react-spring version on a legacy React app (< 19.0.0) is throwing `Error: Uncaught [TypeError: Cannot read properties of null (reading 'ref')]` cause by this PR https://github.com/pmndrs/react-spring/pull/2373

### What

Patching the ref check so it doesn't try to assign the ref to the props object but passes the ref directly to the element.

### Checklist

- [ x ] Ready to be merged


joshuaellis:
Can you add a changeset please? also is there a way to test this change easily?

dimitris-tsetsonis:
@joshuaellis I'll add the changeset today. I'll try to think if I can add a test but since it needs 2 different versions of React I don't think it's possible. I could add a demo with the failing code if you want.

dimitris-tsetsonis:
@joshuaellis I'm really keen to move this PR. Is there anything I need to do to help move the process?

joshuaellis:
I'll do an experimental release tomorrow, you can use it in a small demo w/ r18 to validate it works and share? StackBlitz would work?

joshuaellis:
I've just pushed a beta – https://www.npmjs.com/package/@react-spring/web/v/10.0.3-beta.0

thanks for the nudge 👍🏼

dimitris-tsetsonis:
> I've just pushed a beta – https://www.npmjs.com/package/@react-spring/web/v/10.0.3-beta.0
> 
> thanks for the nudge 👍🏼

@joshuaellis thank you! Will try to push a demo on StackBlitz as soon as possible

dimitris-tsetsonis:
@joshuaellis 
Working demo with 10.0.3-beta.0 -> https://stackblitz.com/edit/vitejs-vite-3owlkm3h

Broken demo with 10.0.2 -> https://stackblitz.com/edit/vitejs-vite-f34k2mgy

Code is exactly the same only react spring version changing. Code is AI but I made sure it reproduces my issue

dimitris-tsetsonis:
@joshuaellis sorry to be a pain. I added the demos in my previous comment and also it was tested by @romain-trotard on his PR #2396  and he too found the solution here to be working.

We would love if you could review and move it for release!

joshuaellis (approved):

joshuaellis:
Yep checked both your demos & even checked the working one on r19 and LGTM ✅

Issue #2395, opened 2025-09-10
[bug]: useTransition not working with React 18


romain-trotard:
### Which react-spring target are you using?

- [x] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

### What version of react-spring are you using?

10.0.2

### What's Wrong?

Hello :) 

With the change made [here](https://github.com/pmndrs/react-spring/pull/2373/files), the hook `useTransition` does not work anymore when using React in v18.

If returning `null` from the transition function we get the error `Cannot read properties of null (reading 'ref')`.
> See reproduction: https://codesandbox.io/p/devbox/react-spring-ref-pb-forked-dszjxn?workspaceId=ws_HL6gchWTGZas2nUL51SZeG

Otherwise we get the error `Cannot add property ref, object is not extensible`
> See reproduction: https://codesandbox.io/p/devbox/elated-shamir-kftfgn?workspaceId=ws_HL6gchWTGZas2nUL51SZeG

### To Reproduce

- use React 18
- use `useTransition` for example to show/hide a `div`

For example with the following code:

```
function Card() {
  const [show, setShow] = useState(false);

  const transition = useTransition(show, {
    config: { duration: 5000 },
    from: { opacity: 0 },
    enter: { opacity: 1 },
    leave: { opacity: 0 },
  });

  return (
    <div>
      <button type="button" onClick={() => setShow(!show)}>
        Show element
      </button>
      {transition(
        (styles, newShow) =>
          newShow && (
            <animated.div styles={styles}>An element to show</animated.div>
          ),
      )}
    </div>
  );
}
```

I made a codesandbox with the problem https://codesandbox.io/p/devbox/elated-shamir-kftfgn?workspaceId=ws_HL6gchWTGZas2nUL51SZeG

### Expected Behaviour

It should not throw an error.

To fix the first issue the condition here https://github.com/pmndrs/react-spring/pull/2373/files#diff-3dce88b788d6e5008705461509e6c3a027695ca9a198d918b9d9c4800d4d2181R441  should be `if (isLegacyReact && elem)` (and if we want to be 100% as before `if (isLegacyReact && elem && elem.type)`

For the second issue recreate a new object object from `elem?.props` works: `const props = { ...elem?.props }`

Thanks

### Link to repo

[...]

PR #2396, opened 2025-09-10
fix: fix `useTransition` when working with React 18

romain-trotard:
Hello :) 

### Why

This is a PR to fix the issue that I've opened https://github.com/pmndrs/react-spring/issues/2395

### What

There are some problems after the change made here https://github.com/pmndrs/react-spring/pull/2373/files#diff-3dce88b788d6e5008705461509e6c3a027695ca9a198d918b9d9c4800d4d2181R441 when working with React 18

### Checklist

ps: the command `yarn changeset:add` does not seem to work. It seems `yarn changeset` adds the changeset file, should I add it? (I see that some PRs are merged without changeset file)

Not working changeset command

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/hooks/useTransition.tsx
+++ b/packages/core/src/hooks/useTransition.tsx
@@ -436,9 +436,9 @@ export function useTransition(
         const key = is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id
         const isLegacyReact = React.version < '19.0.0'
 
-        const props = elem?.props ?? {}
+        const props = { ...elem?.props }
 
-        if (isLegacyReact) {
+        if (isLegacyReact && elem && elem.type) {
           props.ref = elem.ref
         }
 


packages/core/src/hooks/useTransition.tsx
```diff
@@ -436,9 +436,9 @@ export function useTransition(
         const key = is.str(t.key) || is.num(t.key) ? t.key : t.ctrl.id
         const isLegacyReact = React.version < '19.0.0'
 
-        const props = elem?.props ?? {}
+        const props = { ...elem?.props }
 
-        if (isLegacyReact) {
+        if (isLegacyReact && elem && elem.type) {
```
msalettes, line 441: Sorry but do you know why we also need to check elem.type? It seems to me that `&& elem` should be enough. Am I missing something?
I know there’s already the same check earlier; but i do not understand why it is necessary.

romain-trotard: Yep I think `elem` should be enough. But to be completely iso with the previous code I put `elem.type`.

Maybe someone knowing better the codebase can tell why there is the condition `elem && elem.type` a little further in the codebase

dimitris-tsetsonis:
@romain-trotard this should be similar to my PR here https://github.com/pmndrs/react-spring/pull/2394 which has a beta version. Would be create if you could verify it too and we can merge one of the 2 PRs

romain-trotard:
@dimitris-tsetsonis I've just tested the `10.0.3-beta.0` version, and it works well on my project. 
I close my PR ;)

PR #2407, opened 2026-01-20
fix: exitBeforeEnter in useTransition doesn't work when used with trail (#1868)

### LINKED ISSUE #1868: [bug]: exitBeforeEnter in useTransition doesn't work when used with trail

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

9.4.2

### What's Wrong?

When using `useTransition` with both the `trail` and `exitBeforeEnter` option set, the spring freezes after the leave transitions are done. When uncommenting either, it works.

### To Reproduce

Provide `useTransition` with both a `trail` and a `exitBeforeEnter` option.

Checkout the code sandbox: https://codesandbox.io/s/6-loading-spinner-forked-edl8d3?file=/src/SplitWords.js 

When clicking next, the words animate out, but never animate in. If you uncomment `trail` or `exitBeforeEnter` and refresh, it works.

### Expected Behaviour

I would expect the words to animate out with a trail. When there are all animated out, I expect the enter transition of the updated words to start.

### Link to repo

https://codesandbox.io/s/6-loading-spinner-forked-edl8d3?file=/src/SplitWords.js

hiebj:
### Why

Fixes #1868

### What

When `trail` and `exitBeforeEnter` are used together, only the last of the leave transitions was being removed from `exitingTransitions`. (See useTransition.tsx:330) On re-render, the other transitions which remained in `exitingTransitions` meant enter animations would not play.

Instead I am now clearing each transition as it completes. Essentially I just moved the `delete()` statement into the `t.phase == TransitionPhase.LEAVE` block so that it runs for every leaving transition rather than waiting for idle.

I also added a new test. To verify the fix you can pull this commit, remove the fix, and run the test. It will fail. With the fix in place, it passes. No other tests fail and my visual tests in my app show that things still work.

### Checklist

- [x] Documentation updated
- [x] Demo added
- [x] Ready to be merged

I don't think documentation needs to be updated or a demo needs to be added for this bugfix

### DIFF
--- a/.changeset/gentle-tomatoes-nail.md
+++ b/.changeset/gentle-tomatoes-nail.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/core': patch
+---
+
+fix: exitBeforeEnter in useTransition doesn't work when used with trail
--- a/packages/core/src/hooks/useTransition.test.tsx
+++ b/packages/core/src/hooks/useTransition.test.tsx
@@ -195,6 +195,36 @@ describe('useTransition', () => {
 
     expect(rendered).toEqual([1])
   })
+
+  it('should work with both exitBeforeEnter and trail together', async () => {
+    const props = {
+      from: { t: 0 },
+      enter: { t: 1 },
+      leave: { t: 0 },
+      exitBeforeEnter: true,
+      trail: 100,
+    }
+
+    // Start with two items
+    update([0, 1], props)
+    expect(rendered).toEqual([0, 1])
+
+    global.mockRaf.step()
+
+    // Replace with two new items - the old ones should leave first with trail
+    update([2, 3], props)
+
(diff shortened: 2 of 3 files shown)


hiebj:
Can I get a 👀  on this?

hiebj:
@joshuaellis ? Is there another maintainer who can look at this?

I have read many of your comments so I know you are a busy guy, but this is a 1-line PR and includes a new test. I don't know if I can make it much easier to review.

I am planning to contribute more fixes for some of your open issues if I can get this one across, but if they each take a month for review it's going to take a while!

joshuaellis (approved):
Thanks for the nudge, I missed this notification.

PR #2419, opened 2026-05-12
fix(shared): use border box for element resize

### LINKED ISSUE #2288: [bug]: `useResize` uses content box instead of border box

- [X] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

9.7.3

### What's Wrong?

`useResize` returns the size of the content box, without padding or borders. It should use the border box instead.

https://github.com/pmndrs/react-spring/blob/1e4406e322f878141718737638bb6e96e717dd63/packages/shared/src/dom-events/resize/resizeElement.ts#L42-L44

```diff
-   if (observer) {
-     observer.observe(target)
-   }
+   observer?.observe(target, { box: 'border-box' })
```

https://github.com/pmndrs/react-spring/blob/1e4406e322f878141718737638bb6e96e717dd63/packages/shared/src/dom-events/resize/resizeElement.ts#L7-L10

```diff
 const handleObservation = (entries: ResizeObserverEntry[]) =>
-  entries.forEach(({ target, contentRect }) => {
-    return resizeHandlers.get(target)?.forEach(handler => handler(contentRect))
+  entries.forEach(({ target, borderBoxSize }) => {
+    const handlers = resizeHandlers.get(target)
+    const { blockSize, inlineSize } = borderBoxSize[0]
+    const { width, height } = !getComputedStyle(target).getPropertyValue('writing-mode')?.startsWith('vertical-')
+      ? { width: inlineSize, height: blockSize }
+      : { width: blockSize, height: inlineSize }
+    for (const handler of handlers) handler({ width, height })
   })
```

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe#border-box)

### To Reproduce

observe an element that has padding.

### Expected Behaviour

padding should be included in the returned size.

### Link to repo

https://codesandbox.io/p/sandbox/muddy-hill-sjfzwl

apples-kksk:
## Requested
Fixes #2288 
## Implemented
- Observe elements with { box: "border-box" } when supported.
- Prefer ResizeObserverEntry.borderBoxSize for reported dimensions.
- Preserve fallback behavior for older browsers by falling back to contentRect and to plain observe(target) when observe options are unsupported.
- Map inlineSize/blockSize for vertical and sideways writing modes.
- Added unit coverage for border-box sizing, vertical writing mode mapping, sideways writing mode mapping, contentRect fallback, unsupported observe options fallback, and non-TypeError observe failures.

## Not covered
No changes to window resize handling

## Validated
- git diff --check
- Attempted targeted validation with:
- yarn jest packages/shared/src/dom-events/resize/resizeElement.test.ts --runInBand
- yarn prettier --check packages/shared/src/dom-events/resize/resizeElement.ts packages/shared/src/dom-events/resize/resizeElement.test.ts

Both Yarn commands could not run in this checkout because the Yarn node_modules state file is missing and dependencies are not currently installed.

### DIFF
--- a/.changeset/new-melons-cheat.md
+++ b/.changeset/new-melons-cheat.md
@@ -1,0 +1,5 @@
+---
+'@react-spring/shared': patch
+---
+
+Fix element resize measurements to use border-box dimensions when available.
--- a/packages/shared/src/dom-events/resize/resizeElement.test.ts
+++ b/packages/shared/src/dom-events/resize/resizeElement.test.ts
@@ -46,7 +46,7 @@ describe('resizeElement', () => {
           target,
           borderBoxSize: [{ inlineSize: 120, blockSize: 80 }],
           contentRect: { width: 100, height: 60 },
-        } as ResizeObserverEntry,
+        } as unknown as ResizeObserverEntry,
       ],
       {} as ResizeObserver
     )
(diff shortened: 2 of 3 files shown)


joshuaellis:
Thanks for digging into this @apples-kksk — your diagnosis and the writing-mode handling were spot on, and they directly informed #2432.

A few reasons I went with a fresh PR rather than landing this one as-is:

- **Test framework**: this repo migrated to Vitest in real Chromium a while back, so the Jest-based mocks needed rewriting against the real `ResizeObserver` API.
- **`box: 'border-box'` try/catch fallback**: the option has been supported in every evergreen browser since 2020, so I dropped it to keep the surface area small.
- **Changeset wording**: tweaked to describe the user-visible behaviour change.

Really appreciate the work — flagging the writing-mode case in particular saved a regression. 🙏

PR #2535, opened 2026-06-09
fix(animated): WeakMap fallback for non-extensible component caching

### LINKED ISSUE #2533: animated() throws 'cannot add a new property' on non-extensible React Native host components (Hermes)
## Bug

`@react-spring/animated`'s `createHost` caches animated wrappers by writing directly to the component object:

```js
Component[cacheKey] = withAnimated(Component, hostConfig)
```

On **Hermes**, React Native host components (`View`, `Text`, `Image`) become **non-extensible** after their first JSX render. In strict mode this throws `TypeError: cannot add a new property`, which crashes the module initialization.

## Reproduction

- React Native with Hermes
- Metro config: `inlineRequires: true` + `experimentalImportSupport: true`
- Any component that uses `@react-spring/native` loaded lazily (inside a render)

With inlined requires, `@react-spring/native` is first required during a component render — by that point Hermes has already rendered the host components via JSX elsewhere and locked their shapes. `createHost` runs at module scope, iterates `{ View, Text, Image }`, and throws on the first `Component[cacheKey] = ...` assignment.

Metro's `guardedLoadModule` swallows the error and returns `undefined` for the module, causing a subsequent `TypeError: Cannot read property 'useSpring' of undefined`.

## Call stack (from device)

```
animated          react-spring_animated.development.cjs:356
<anonymous>       react-spring_animated.development.cjs:365
eachProp          react-spring_shared.development.cjs:120
createHost        react-spring_animated.development.cjs:361
<global>          react-spring_native.development.cjs:90
loadModuleImplementation  require.js:285
guardedLoadModule
```

## Fix

Use the direct property write as the fast path, catch the `TypeError`, and fall back to a module-level `WeakMap`:

```ts
const fallbackCache = new WeakMap<object, any>()

// inside animated():
let cached = Component[cacheKey] ?? fallbackCache.get(Component)
if (!cached) {
  cached = withAnimated(Component, hostConfig)
  try {
    Component[cacheKey] = cached
  } catch {
    // non-extensible component (e.g. Hermes host component)
  }
  fallbackCache.set(Component, cached)
}
Component = cached
```

PR with fix: https://github.com/AndreiCalazans/react-spring/pull/new/fix/non-extensible-component-cache

AndreiCalazans:
Fixes #2533

## Problem

`createHost` caches animated wrappers via `Component[cacheKey] = ...`. On Hermes, React Native host components (`View`, `Text`, `Image`) become non-extensible after their first JSX render — writing a new property throws `TypeError: cannot add a new property` in strict mode.

This crashes `@react-spring/native`'s module initialization when loaded lazily (e.g. with Metro `inlineRequires: true`), because `createHost` runs at module scope iterating over those exact primitives.

## Fix

Try the direct write first (fast path, no change for extensible components). On failure, fall back to a module-level `WeakMap`.

```diff
+const fallbackCache = new WeakMap<object, any>()

 const animated: WithAnimated = (Component: any) => {
   if (is.str(Component)) {
     Component = animated[Component] || (animated[Component] = withAnimated(Component, hostConfig))
   } else {
-    Component = Component[cacheKey] || (Component[cacheKey] = withAnimated(Component, hostConfig))
+    let cached = Component[cacheKey] ?? fallbackCache.get(Component)
+    if (!cached) {
+      cached = withAnimated(Component, hostConfig)
+      try {
+        Component[cacheKey] = cached
+      } catch {
+        // non-extensible component — store in WeakMap fallback
+      }
+      fallbackCache.set(Component, cached)
+    }
+    Component = cached
   }
 }
```

## Context

This was found while enabling Metro's `inlineRequires: true` + `experimentalImportSupport: true` in a large React Native monorepo. A local Yarn patch on `@react-spring/animated@10.0.3` with this exact fix is already shipping in production while we wait for an upstream release: https://coinbase.ghe.com/consumer/react-native/pull/new/acalazans/metro-inline-requires-enable

### DIFF
--- a/.changeset/non-extensible-component-cache.md
+++ b/.changeset/non-extensible-component-cache.md
@@ -1,0 +1,9 @@
+---
+'@react-spring/animated': patch
+---
+
+Fix `cannot add a new property` crash when wrapping non-extensible React Native host components.
+
+`createHost` cached animated wrappers by writing directly to the component object via `Component[Symbol.for('AnimatedComponent')] = ...`. On Hermes (React Native), host components like `View`, `Text`, and `Image` become non-extensible after their first JSX render, causing a `TypeError` in strict mode.
+
+The fix attempts the direct write first (fast path, no change for extensible components) and falls back to a module-level `WeakMap` when the write is rejected.
--- a/packages/animated/src/createHost.test.ts
+++ b/packages/animated/src/createHost.test.ts
@@ -1,5 +1,4 @@
 import { createHost } from './createHost'
-import { withAnimated } from './withAnimated'
 import { AnimatedObject } from './AnimatedObject'
 
 const hostConfig = {
@@ -10,7 +9,9 @@ const hostConfig = {
 
 describe('createHost', () => {
   it('caches the animated wrapper on the component itself', () => {
-    function MyComponent() {}
+    function MyComponent() {
+      return null
+    }
     const { animated } = createHost({ MyComponent }, hostConfig)
 
     const A = animated(MyComponent)
@@ -21,7 +22,9 @@ describe('createHost', () => {
   })
 
   it('falls back to a WeakMap when the component is non-extensible', () => {
-    function MyComponent() {}
+    function MyComponent() {
+      return null
+    }
     Object.preventExtensions(MyComponent)
 
     // Should not throw even though MyComponent is non-extensible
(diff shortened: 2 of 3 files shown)


joshuaellis (changes requested):
it'd be great to get a failing test for this & a changeset.

joshuaellis (dismissed):

joshuaellis:
Looks like there's some TS errors that need to be solved.

AndreiCalazans:
@joshuaellis thanks for flagging the CI failure. Fixed.

joshuaellis (approved):

PR #2547, opened 2026-06-19
fix: cache output number parsing in createStringInterpolator

literalpie:
### Why

Starting in version 10.1.0, this code is very slow when showing a chart with a large number of points.

### What

Instead of calculating the value in every iteration of the loop, the numbers are only calculated once.

### Checklist

- [x] Ready to be merged

### DIFF
--- a/.changeset/clever-islands-dig.md
+++ b/.changeset/clever-islands-dig.md
@@ -1,0 +1,13 @@
+---
+'@react-spring/shared': patch
+'@react-spring/animated': patch
+'@react-spring/core': patch
+'@react-spring/mock-raf': patch
+'@react-spring/parallax': patch
+'@react-spring/rafz': patch
+'@react-spring/types': patch
+'@react-spring/three': patch
+'@react-spring/web': patch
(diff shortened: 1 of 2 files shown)


literalpie:
heads up: the contributing.md file implies that only maintainers need to worry about changesets, and says
> With `react-spring` we bump every package to the same version, so flag them all

I'm guessing "maintainers only" isn't right based on the above comment, and I'm not sure if "flag them all" is right, since I see other PRs that don't flag all packages

joshuaellis (approved):

joshuaellis:
Thanks for the contribution, i've done a follow up PR for the contributing recommendations – #2549, thanks for flagging.

literalpie:
Any chance this could be backported to 10.x 🙏? Or can you share if there's any timeline on when 11.x will be stable?

joshuaellis:
No timeline, very much down time development. I'll look at back porting today.

literalpie:
The backport helped us a lot. Thanks!

PR #2548, opened 2026-06-20
fix(core): pass an AnimationResult to SpringValue onChange

### LINKED ISSUE #2183: [bug]: Incorrect result type in SpringValue `onChange` handler

- [X] `@react-spring/web`
- [X] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

9.7.3

### What's Wrong?

The `onChange` option in the `SpringValue` constructor indicates that the argument is of type `AnimationResult<SpringValue<number>>`

CleanShot 2023-08-03 at 13 48 53@2x

Thus, `result.value` is indicated to be of type `number`, and logging this should display the animated value. However, it actually prints undefined as the spring animates:

CleanShot 2023-08-03 at 13 49 39@2x

If I instead log the entire `result`, I see the number.

### To Reproduce

```ts
import { SpringValue } from '@react-spring/web'

const spring = new SpringValue(0, {
 onChange(result) {
  console.log(result.value)
 }
})

spring.start(1)

```

### Expected Behaviour

I expect to see the value on `result.value`, which TypeScript indicates is a `number`

### Link to repo

https://codesandbox.io/s/vigorous-mirzakhani-qrrwj8?file=/src/App.tsx

patrickwehbe:
### Why

Fixes #2183. The `SpringValue`-level `onChange` handler is typed to receive an `AnimationResult`, so `result.value` should be the animated value. At runtime it was handed the raw value instead, so `result.value` was `undefined` for the whole animation (reading `.value` off a number). Logging the argument directly showed the number, which is what the issue reports.

The fix itself was diagnosed by @joshuaellis in the issue thread; this PR implements it and adds the test coverage he suggested.

### What

`SpringValue._onChange` now wraps the value in a result before calling the handlers, mirroring how `onStart` and `onRest` dispatch theirs and matching the value the `Controller`-level `onChange` already provides:

```ts
const result = getFinishedResult(value, false) // { value, finished: false, cancelled: false }
```

Both `onChange` call sites (the active `animation.onChange` and `defaultProps.onChange`) receive `result`. `getFinishedResult` was already imported, so there are no new imports.

`super._onChange(value, idle)` deliberately keeps passing the raw value. That call raises the internal `change` event that the animated tree and the `Controller` subscribe to, and those observers read the value itself, not a result wrapper. Only the user-facing `onChange` was wrong.

Tests, in `packages/core/src/SpringValue.test.ts`:

- Added a case asserting the `onChange` argument is an `AnimationResult` (never the bare value) and that `result.value` equals the spring's current value on every frame, ending on the target value. Reverting the source change makes this fail with the value arriving as a plain `number`, reproducing the bug.
- Updated the three existing assertions that read the first `onChange` argument as the raw value to read `result.value` instead.

The type-level contract is already pinned by the `#2183` guard in `SpringValue.test-d.ts`, which keeps passing since the type was never wrong.

### Checklist

- [ ] Documentation updated
- [ ] Demo added
- [x] Ready to be merged

### DIFF
--- a/.changeset/fix-springvalue-onchange-result.md
+++ b/.changeset/fix-springvalue-onchange-result.md
@@ -1,0 +1,7 @@
+---
+'@react-spring/core': patch
+---
+
+fix(core): pass an AnimationResult to SpringValue `onChange`
+
+The `SpringValue`-level `onChange` was called with the raw value instead of an `AnimationResult`, so `result.value` was `undefined` mid-animation even though the type advertises it as the value. `onChange` now receives `{ value, finished: false, cancelled: false }`, matching `onStart`/`onRest` and the `Controller`-level `onChange`. The internal `change` event the animated tree subscribes to still emits the raw value.
--- a/packages/core/src/SpringValue.test.ts
+++ b/packages/core/src/SpringValue.test.ts
@@ -70,11 +70,40 @@ describe('SpringValue', () => {
       onChange,
     })
     await global.advanceUntilIdle()
+    // `onChange` receives an AnimationResult (not the raw value); `result.value`
+    // holds the array. See #2183.
     expect(onChange.mock.calls.slice(-1)[0]).toEqual([
-      spring.animation.to,
+      { value: spring.animation.to, finished: false, cancelled: false },
       spring,
     ])
     expect(global.getFrames(spring)).toMatchSnapshot()
+  })
+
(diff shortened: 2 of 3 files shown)


joshuaellis (approved):
Thanks for the contribution!

PR #2557, opened 2026-08-16
fix(core): propagate immediate flag to to-array chain steps

### LINKED ISSUE #2204: [bug]: Immediate does not work when to is an array

- [X] `@react-spring/web`
- [X] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

9.7.3 amd 10.0.0-alpha.1

### What's Wrong?

When I have an array for the `to`-prop in order to chain animations and then set `immediate` to true, the animation runs normally and does not jump to the goal value.

### To Reproduce

1. use `useSpring` with an array for  the `to`-prop
2. set `immediate: true`

### Expected Behaviour

The animation chain should jump right to the end of the last step without animating anything

### Link to repo

https://codesandbox.io/s/clever-visvesvaraya-w5f9lg?file=/src/App.js

ErfanBagheri404:
## Summary

When `immediate: true` is set with an array `to` prop, each step in the chain should start immediately. Previously the flag was only checked in `SpringValue.start()`, but when `to` is an array, `SpringValue.start()` delegates to `runAsync()`, which did not propagate `immediate` to each step.

This caused the chain to animate each step sequentially instead of jumping immediately to the final value.

## Changes

- In `runAsync.ts`: propagate `props.immediate` to each step in the `to` array chain when it is not explicitly set on the step itself

## Test plan

- Added regression test in `Controller.test.ts`
- Changeset added: `@react-spring/core: patch`

## Links

- Fixes pmndrs/react-spring#2204

### DIFF
--- a/.changeset/immediate-to-array-chain.md
+++ b/.changeset/immediate-to-array-chain.md
@@ -1,0 +1,5 @@
+---
+"@react-spring/core": patch
+---
+
+Propagate the `immediate` flag to every step of a `to` array chain. Previously, `immediate: true` was ignored when combined with an array `to` prop, so the animation chain would animate each step sequentially instead of jumping instantly to the final value.
--- a/packages/core/src/Controller.test.ts
+++ b/packages/core/src/Controller.test.ts
@@ -719,6 +719,22 @@ describe('Controller', () => {
       expect(onStart).toHaveBeenCalled()
       expect(onRest).toHaveBeenCalled()
     })
+
+    // Regression test for https://github.com/pmndrs/react-spring/issues/2204
+    it('applies `immediate` to every step of a `to` array chain', async () => {
+      const ctrl = new Controller<{ t: number }>({ t: 0 })
+      const { t } = ctrl.springs
+
+      ctrl.start({
+        to: [{ t: 1 }, { t: 2 }],
+        immediate: true,
+      })
+
+      await global.advanceUntilIdle()
+
+      // With `immediate`, the chain must jump straight to the last step's goal.
+      expect(t.get()).toBe(2)
+    })
   })
 
   // Regression test for https://github.com/pmndrs/react-spring/issues/2208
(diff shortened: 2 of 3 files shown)


joshuaellis:
Thanks for the PR. We're going to close this one, but I want to explain the reasoning clearly since the underlying request is valid.

**The short version:** `immediate` not propagating to `to`-array chain steps is intentional, not a bug. The supported way to apply it across a whole chain is the object syntax:

```js
ctrl.start({ to: [{ t: 1 }, { t: 2 }], default: { immediate: true } })
```

This already works today and is consistent with how the async-script path (`to` as a function) behaves. See the note in `helpers.ts` — `immediate`, `pause`, and `cancel` are deliberately excluded from the implicit default-prop set and must be passed via `default: {}`.

**Why we can't merge this specific change:**

1. **It mutates caller-owned objects.** The loop writes `queueProps.immediate = true` directly onto elements of the user's `to` array before they're spread. With memoized or module-scope step arrays (the common case), that flag persists on the reference and leaks into unrelated animations that never requested `immediate`.

2. **The test doesn't verify the fix.** `advanceUntilIdle()` drives the chain to completion either way, so `expect(t.get()).toBe(2)` passes whether the value jumped instantly or animated over many frames. It passes against `next` unpatched, so it wouldn't catch a regression.

3. **It only handles `immediate === true`.** The prop also accepts a predicate function and an array of keys; those forms would still animate step-by-step, making the behavior inconsistent.

4. **Semantics + changeset mismatch.** Changing documented default-prop propagation is a behavioral change, not a `patch`. If we did want bare `immediate` to propagate, it would need to be a deliberate, consistent change across both the array and script paths and all three `immediate` shapes — routed through the existing `defaultProps`/`getDefaultProps` layer rather than mutating step objects.

That said — enough people keep hitting this that I don't think the answer is "read the docs." Bare `immediate: true` silently not applying to chain steps is a genuine footgun, and we should explore how to support it properly rather than closing the door. I'm keeping the original issue open to scope the proper fix (consistent handling across array and script `to`, all three `immediate` shapes, no caller mutation). This PR is closed only because the implementation can't ship as-is, not because the goal is wrong.

PR #2558, opened 2026-08-16
fix(animated): use WeakMap cache in createHost

### LINKED ISSUE #2533: animated() throws 'cannot add a new property' on non-extensible React Native host components (Hermes)

ErfanBagheri404:
## Summary

On React Native (Hermes), host components (`View`, `Text`, `Image`) become **non-extensible** after their first JSX render. `createHost` previously stashed the animated wrapper directly on the component:

```ts
Component[cacheKey] || (Component[cacheKey] = withAnimated(Component, hostConfig))
```

With Metro `inlineRequires` + `experimentalImportSupport`, `@react-spring/native` can first be required during a render — by which point Hermes has locked the host component shapes — so `createHost` threw `TypeError: cannot add a new property`, and Metro's `guardedLoadModule` swallowed it returning `undefined`, cascading into `Cannot read property 'useSpring' of undefined`.

## Changes

- Replaced the per-component `Symbol.for('AnimatedComponent')` property with a module-level `WeakMap` cache in `createHost`.
- The original component is never mutated, so non-extensible host components are safe.
- Added a regression test (`createHost.test.ts`) that freezes a fake host component and asserts `animated()` neither throws nor returns the same reference.

## Test plan

- `packages/animated/src/createHost.test.ts` (new)
- Changeset: `@react-spring/animated: patch`

## Links

- Fixes pmndrs/react-spring#2533

### DIFF
--- a/packages/animated/src/createHost.test.ts
+++ b/packages/animated/src/createHost.test.ts
@@ -57,6 +57,7 @@ describe('createHost', () => {
     const { animated } = createHost({ NamedComponent }, hostConfig)
 
     const A = animated(NamedComponent)
+
     expect(A.displayName).toBe('Animated(NamedComponent)')
   })
 
--- a/packages/animated/src/createHost.ts
+++ b/packages/animated/src/createHost.ts
@@ -19,7 +19,7 @@ type WithAnimated = {
   [key: string]: any
 }
 
-// For storing the animated version on the original component
+/** Module-level cache so animated wrappers survive across createHost calls */
 const cacheKey = Symbol.for('AnimatedComponent')
 
 // Fallback cache for component objects that are non-extensible (e.g. React


joshuaellis:
WeakMap swap looks right.

CI is failing on formatting only. The else block in createHost.ts is over-indented so oxfmt fails. Run pnpm format and push.

Optional: the nested if/else can be one branch.

```ts
} else {
  let cached = animatedComponentCache.get(Component)
  if (!cached) {
    cached = withAnimated(Component, hostConfig)
    animatedComponentCache.set(Component, cached)
  }
  Component = cached
}
```

ErfanBagheri404:
Done — formatted and pushed. The else branch now uses your suggested shape (let cached + single set), and the branch is a single commit on `next`.

joshuaellis:
you targetted the wrong base branch, if you check `next` (without using an LLM) you'll see react-native was removed, you need to focus on the v10 branch for this imo.

joshuaellis (changes requested):
needs correct rebase target branch

ErfanBagheri404:
Hello, imo you may now close the PR and resolve the issue (without using an LLM) Alternatively, I'm happy to address the requested changes.
Thanks!

PR #2559, opened 2026-08-16
fix(core): prevent infinite loop when skipAnimation + loop (#2409)

### LINKED ISSUE #2409: [bug]: Tab crash with skipAnimation and loop

- [x] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

10.0.3

### What's Wrong?

When combining a SpringValue that has `loop: true` and `Globals.assign({skipAnimation:true})`, there is an infinite loop that crashes the tab.

When setting immediate: true, this infinite loop runs every frame (with the request animation frame loop).

### To Reproduce

This minimal code is enough to lock the main thread at 100% usage.

```js
import { Globals, SpringValue } from "@react-spring/web";

Globals.assign({ skipAnimation: true });

new SpringValue(0, {
  to: 1,
  loop: true,
});
```

### Expected Behaviour

If we are setting skipAnimation or immediate, I think the loop parameter should be ignored to avoid this infinite loop.

### Link to repo

https://codesandbox.io/p/sandbox/tender-scott-5k6l8w

ErfanBagheri404:
## Summary

When `skipAnimation: true` (via `Globals.assign({ skipAnimation: true })`) is combined with `loop: true`, the tab hangs at 100% CPU. This reproduces with:

```js
import { Globals, SpringValue } from "@react-spring/web";
Globals.assign({ skipAnimation: true });
new SpringValue(0, { to: 1, loop: true });
```

## Root cause

With `skipAnimation` enabled, `SpringValue._resume()` calls `this.finish()` synchronously instead of deferring to the frameloop. The promise resolves immediately, and the `.then()` loop callback re-enters via `_update() -> _start() -> _resume() -> finish()`. Since nothing yields to the event loop, this recurses infinitely and freezes the tab.

## Changes

- In `SpringValue.ts`, skip the loop re-trigger in the `.then()` callback when `G.skipAnimation` is enabled.
- The spring settles at its goal value instead of looping.

## Test plan

- New regression test in `SpringValue.test.ts` (loop + skipAnimation settles immediately at goal, becomes idle).

## Links

- Fixes pmndrs/react-spring#2409

### DIFF
--- a/packages/core/src/SpringValue.test.ts
+++ b/packages/core/src/SpringValue.test.ts
@@ -757,6 +757,21 @@ function describeLoopProp() {
       expect(spring.idle).toBeFalsy()
     })
 
+    it('settles immediately when skipAnimation is enabled (#2409)', async () => {
+      const spring = new SpringValue(0)
+      Globals.assign({ skipAnimation: true })
+      spring.start(1, { loop: true })
+
+      await global.advanceUntilIdle()
+
+      // With skipAnimation, the spring must settle at its goal instead of
+      // re-entering the loop synchronously (which used to hang the tab).
+      expect(spring.get()).toBe(1)
+      expect(spring.idle).toBeTruthy()
+
+      Globals.assign({ skipAnimation: false })
+    })
+
     it('can be combined with the "reverse" prop', async () => {
       const spring = new SpringValue(0)
       spring.start(1, { config: { duration: frameLength * 3 } })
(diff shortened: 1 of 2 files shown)


joshuaellis:
Thanks for tackling #2409. Dug into this on `next` — the fix targets the wrong case.

`skipAnimation + loop` doesn't freeze on `next`. the new test passes with and without the fix, so it doesn't cover a regression.

`immediate + loop` is the broken behaviour (also named in #2409).

PR #2560, opened 2026-08-16
fix(shared): match add/removeEventListener options in onScroll cleanup (#2384)

### LINKED ISSUE #2384: [bug]: event listeners cleanup fail (potential memory leak)

- [x] `@react-spring/web`
- [ ] `@react-spring/three`
- [x] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

10.0.1

### What's Wrong?

When using `useScroll` event listeners are not being properly cleaned-up when the component unmounts.

I narrowed down the issue to the following two files under `@react-spring/shared` which are not correctly matching `addEventListener` with `removeEventListener`:

- [`./src/dom-events/scroll/index.ts`](https://github.com/pmndrs/react-spring/blob/next/packages/shared/src/dom-events/scroll/index.ts)
- [`./src/dom-events/resize/resizeWindow.ts`](https://github.com/pmndrs/react-spring/blob/next/packages/shared/src/dom-events/resize/resizeWindow.ts)

Both implementation are ignoring `EventListenerOptions` while, according to the documentation (see [Matching event listeners for removal](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal)):
- At least `capture` option must match across `add` and `remove` calls.
- In any case, the recommendation is to use all the same options/args in both calls, because not al browsers behave the same in this regard.

We have automatic tests for event listeners cleanup, and only `useScroll` is failing in the entire codebase:

```
Error: UNMATCHED EVENT LISTENERS:
    - SOURCE: window
    - ADD: 4
    - REMOVE: 4
    - UNMATCHED: 4


ADD CALLS:
    - TYPE: resize
    - LISTENER: (REF ID: 0) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: {"passive":true}


    - TYPE: scroll
    - LISTENER: (REF ID: 0) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: {"passive":true}


    - TYPE: resize
    - LISTENER: (REF ID: 1) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: {"passive":true}


    - TYPE: scroll
    - LISTENER: (REF ID: 1) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: {"passive":true}


REMOVE CALLS:
    - TYPE: scroll
    - LISTENER: (REF ID: 0) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: undefined


    - TYPE: resize
    - LISTENER: (REF ID: 0) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: undefined


    - TYPE: scroll
    - LISTENER: (REF ID: 1) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: undefined


    - TYPE: resize
    - LISTENER: (REF ID: 1) () => { containerHandlers?.forEach((handler) => handler.advance()); return true; }
    - OPTIONS: undefined
```

### To Reproduce

Just use `useScroll` and check for DOM listeners in the inspector. Results might vary depending on browser/environment.

### Expected Behaviour

`useScroll` to properly cleanup event listeners.

### Link to repo

N/A.

**I would be happy to open a PR to improve/fix this.**

ErfanBagheri404:
## Summary

`useScroll` (via `onScroll` in `@react-spring/shared`) leaks event listeners on cleanup.

`addEventListener` was called with `{ passive: true }`:

```ts
window.addEventListener('resize', listener, { passive: true })
target.addEventListener('scroll', listener, { passive: true })
```

but `removeEventListener` was called without any options:

```ts
getTarget(container).removeEventListener('scroll', listener)
window.removeEventListener('resize', listener)
```

Per the [DOM spec on matching event listeners](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal), the `capture` flag must match between add and remove for the listener to be removed. While `{ passive: true }` happens to default `capture` to `false`, strict environments that track the options object verbatim fail to match, leaving the listener attached — `useScroll` leaks a scroll + resize listener per component mount/unmount.

## Changes

- `packages/shared/src/dom-events/scroll/index.ts`: extract a shared `passiveOptions` object used by **both** `addEventListener` and `removeEventListener` calls (scroll + resize + per-container `onResize` path), so the options always match.
- `packages/shared/src/dom-events/scroll/onScroll.test.ts`: regression test asserting `removeEventListener` is invoked with options equal to those passed to `addEventListener`.

## Test plan

- `npx vitest run packages/shared/src/dom-events/scroll/onScroll.test.ts` — passes (mock throws on unmatched remove, none thrown)

Fixes pmndrs/react-spring#2384

### DIFF
--- a/.changeset/fix-onscroll-listener-cleanup.md
+++ b/.changeset/fix-onscroll-listener-cleanup.md
@@ -1,0 +1,7 @@
+---
+'@react-spring/shared': patch
+---
+
+fix(shared): match add/removeEventListener options in onScroll cleanup
+
+`onScroll` added its scroll and resize listeners with `{ passive: true }` but removed them with no options. A shared options object is now passed to both the `addEventListener` and `removeEventListener` calls so they always match, preventing `useScroll` from leaking a scroll and resize listener per mount in environments that match listeners on the options object.
--- a/packages/shared/src/dom-events/scroll/index.ts
+++ b/packages/shared/src/dom-events/scroll/index.ts
@@ -12,6 +12,11 @@ export type OnScrollOptions = {
   container?: HTMLElement
 }
 
+// Shared options used for both addEventListener and removeEventListener.
+// Must be passed to both calls for correct listener matching in strict
+// environments, otherwise removeEventListener silently no-ops (#2384).
+const passiveOptions: AddEventListenerOptions = { passive: true }
+
 const scrollListeners = new WeakMap<Element, () => boolean>()
 const resizeListeners = new WeakMap<Element, VoidFunction>()
 const onScrollHandlers = new WeakMap<Element, Set<ScrollHandler>>()
@@ -62,7 +67,7 @@ export const onScroll = (
      * Add resize handlers so we can correctly calculate the
      * scroll position on changes
      */
-    window.addEventListener('resize', listener, { passive: true })
+    window.addEventListener('resize', listener, passiveOptions)
 
     if (container !== document.documentElement) {
       resizeListeners.set(container, onResize(listener, { container }))
@@ -71,7 +76,7 @@ export const onScroll = (
     /**
      * Add the actual scroll listener
      */
-    target.addEventListener('scroll', listener, { passive: true })
+    target.addEventListener('scroll', listener, passiveOptions)
   }
 
   /**
@@ -103,10 +108,12 @@ export const onScroll = (
     scrollListeners.delete(container)
 
     if (listener) {
-      getTarget(container).removeEventListener('scroll', listener)
-      window.removeEventListener('resize', listener)
+          // Match the options used by addEventListener, otherwise the listener
+          // is not removed in strict environments (see #2384).
+          getTarget(container).removeEventListener('scroll', listener, passiveOptions)
+          window.removeEventListener('resize', listener, passiveOptions)
 
-      resizeListeners.get(container)?.()
-    }
+          resizeListeners.get(container)?.()
+        }
   }
 }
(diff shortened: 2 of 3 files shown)


joshuaellis (changes requested):
you've got your two other PR commits in this. I think realistically you wanted to stack these PRs but to be honest, they should be made in isolation with each PR a single commit aimed at `next`. Please review all your PRs and amend accordingly before they can be reviewed.

ErfanBagheri404:
Rebased and isolated all 5 PRs onto `next` — each branch now carries only its own single commit (no stacked siblings):

- #2557 immediate-to-array-chain
- #2558 createhost-weakmap-cache
- #2559 skip-animation-loop-crash
- #2560 onscroll-listener-cleanup
- #2561 number-regex-negative-decimals

Force-pushed. Each is now a single isolated commit against `next`.

joshuaellis:
Source fix in `index.ts` is correct — sharing one `passiveOptions` object makes add/remove match, leak gone.

Two blockers:

**Test fails here.** Tests run in Vitest browser mode (real Chromium), so the mock `container` hits the real `ResizeObserver.observe()` and throws before cleanup:

```
TypeError: Failed to execute 'observe' on 'ResizeObserver': parameter 1 is not of type 'Element'.
```

See `resizeElement.test.ts` in the same folder for the pattern — real `document.createElement('div')`. A real element also exposes a bad assertion: only `scroll` is added to the container (`resize` goes on `window`), so `toHaveBeenCalledTimes(2)` won't hold.

**Formatting.** `oxfmt --check` fails on both files (cleanup block in `index.ts` is mis-indented). Run the formatter.

joshuaellis (approved):

PR #2561, opened 2026-08-16
fix(shared): parse negative decimals without integer part in numberRegex (#2397)

### LINKED ISSUE #2397: [bug]: Error when transitioning particular lab colors

- [x] `@react-spring/web`
- [ ] `@react-spring/three`
- [ ] `@react-spring/native`
- [ ] `@react-spring/konva`
- [ ] `@react-spring/zdog`

10.0.2

### What's Wrong?

When transitioning an element from the color `oklch(0.81 0 255 / 1)` to `oklch(0.4 0.2639 271.35 / 1)`, an error is thrown:

```
Uncaught (in promise) Error: The arity of each "output" value must be equal
    at react-spring_shared.modern.mjs:708:15
    at Array.map (<anonymous>)
    at react-spring_shared.modern.mjs:706:25
    at Array.map (<anonymous>)
    at createStringInterpolator2 (react-spring_shared.modern.mjs:705:37)
    at createInterpolator (react-spring_shared.modern.mjs:476:12)
    at _AnimatedString.reset (react-spring_animated.modern.mjs:103:42)
    at SpringValue._start (react-spring_core.modern.mjs:1168:24)
    at SpringValue._merge (react-spring_core.modern.mjs:1082:12)
    at onStart (react-spring_core.modern.mjs:348:17)
    at onResume (react-spring_core.modern.mjs:335:9)
    at react-spring_core.modern.mjs:319:9
    at new Promise (<anonymous>)
    at scheduleProps (react-spring_core.modern.mjs:299:10)
    at SpringValue._update (react-spring_core.modern.mjs:912:12)
    at react-spring_core.modern.mjs:834:25
    at Array.map (<anonymous>)
    at SpringValue.start (react-spring_core.modern.mjs:833:13)
    at react-spring_core.modern.mjs:1516:32
    at Array.map (<anonymous>)
    at flushUpdate (react-spring_core.modern.mjs:1515:56)
    at react-spring_core.modern.mjs:1465:43
    at Array.map (<anonymous>)
    at flushUpdateQueue (react-spring_core.modern.mjs:1465:28)
    at Controller.flush [as _flush] (react-spring_core.modern.mjs:1714:31)
    at Controller.start (react-spring_core.modern.mjs:1380:19)
    at useSprings.useIsomorphicLayoutEffect2 (react-spring_core.modern.mjs:1773:16)
    at Array.forEach (<anonymous>)
    at each (react-spring_shared.modern.mjs:41:29)
```

The error is because the gray color is turned into `lab(77.96% -.0000298023 0)` (either by Chrome or tailwind, I'm not sure), which is then parsed incorrectly in `createStringInterpolator` by the number regex into `[77.96, 0, 0, 0, 0, 298023, 0]` instead of `[77.96, -0.0000298023 0]`.

### To Reproduce

1. `npx create next-app@latest <dir> --yes`
2. `cd <dir>; npm install --save @react-spring/web`
3. In `globals.css`, add to the `:root` block:

```css
  --gray: oklch(0.81 0 255 / 1);
  --blue: oklch(0.4 0.2639 271.35 / 1)
```

4. Change `page.tsx` to:

```tsx
"use client";
import { animated, useSpring } from "@react-spring/web";

export default function Home() {
  const style = useSpring({
    from: { backgroundColor: "var(--gray)" },
    to: { backgroundColor: "var(--blue)" }
  });

  return (
    <animated.div style={style} className="w-10 h-10 rounded-full flex items-center justify-center">
      Hello world
    </animated.div>
  );
}
```

5. Run server, open browser, check console errors

### Expected Behaviour

No error thrown

### Link to repo

https://github.com/JoeEnnever/react-spring-error

ErfanBagheri404:
## Summary

Transitioning between certain CSS color formats throws `"The arity of each output value must be equal"`. The crash happens when a color string contains a decimal without an integer part, such as `-.0000298023` (produced by Chrome/Tailwind for `oklch(0.81 0 255 / 1)` → `lab(77.96% -.0000298023 0)`).

## Root cause

`numberRegex` in `packages/shared/src/regexs.ts` requires at least one digit before the decimal point (`(?:0|[1-9]\d*)`). Input `-.0000298023` is split into fragments `0, 0, 0, 0, 0, 298023` instead of a single token, giving 7 numbers vs the other keyframe's 4 → arity check fails.

## Fix

Replace the regex with one that matches the fractional part independently of having an integer prefix:

```diff
- /[+\-]?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?/g
+ /[+\-]?(?:(?:\d*\.\d+)|\d+)(?:[eE][+\-]?\d+)?/g
```

This correctly handles `-.5`, `-.0000298023`, `+1.5e-3`, and all previously supported formats (verified against 14 test inputs).

## Test plan

- Repro case no longer throws: `createStringInterpolator({output: ['lab(77.96% -.0000298023 0)', 'oklch(0.4 0.2639 271.35 / 1)']})` — works.
- New regression test in `stringInterpolation.test.ts` for lab() colors.
- New test for standalone `-.5px` negative decimals.
- `@react-spring/shared` builds clean (turbo).

Fixes pmndrs/react-spring#2397

### DIFF
--- a/.changeset/fix-number-regex-negative-decimals.md
+++ b/.changeset/fix-number-regex-negative-decimals.md
@@ -1,0 +1,7 @@
+---
+'@react-spring/shared': patch
+---
+
+fix(shared): parse negative decimals without an integer part in numberRegex
+
+`numberRegex` required at least one digit before the decimal point, so a value like `-.0000298023` (emitted by Chrome for some `lab()`/`oklch()` colors) was split into several tokens instead of one. That inflated a keyframe's number count and threw `The arity of each "output" value must be equal` when interpolating between color formats. The regex now matches a leading-dot fraction as a single token while keeping all previously supported number formats.
--- a/packages/shared/src/regexs.ts
+++ b/packages/shared/src/regexs.ts
@@ -1,6 +1,6 @@
 // Problem: https://github.com/animatedjs/animated/pull/102
 // Solution: https://stackoverflow.com/questions/638565/parsing-scientific-notation-sensibly/658662
-export const numberRegex = /[+\-]?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?/g
+export const numberRegex = /[+\-]?(?:(?:\d*\.\d+)|\d+)(?:[eE][+\-]?\d+)?/g
 
 // Covers rgb, rgba, hsl, hsla
 // Taken from https://gist.github.com/olmokramer/82ccce673f86db7cda5e
--- a/packages/shared/src/stringInterpolation.test.ts
+++ b/packages/shared/src/stringInterpolation.test.ts
@@ -38,6 +38,6 @@ it('parses a standalone negative decimal without an integer part', () => {
     output: ['translateY(-.5px)', 'translateY(10px)'],
   })
 
-  expect(() => interpolate(0.5)).not.toThrow()
+  expect(interpolate(0.5)).toBe('translateY(4.75px)')
   expect(interpolate(1)).toBe('translateY(10px)')
 })
--- a/packages/shared/src/stringInterpolation.test.ts
+++ b/packages/shared/src/stringInterpolation.test.ts
@@ -16,3 +16,28 @@ it('interpolates a number-less output value instead of throwing on a null match'
   expect(interpolate(0)).toBe('none')
   expect(interpolate(0.5)).toBe('none')
 })
+
+// https://github.com/pmndrs/react-spring/issues/2397
+// lab() values like `lab(77.96% -.0000298023 0)` contain decimals with no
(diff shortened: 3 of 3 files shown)


joshuaellis (changes requested):
The second test (`parses a standalone negative decimal without an integer part`) passes on `next` without the regex change, so it doesn't guard the fix. `interpolate(1)` hits the keyframe short-circuit and returns `output[1]` verbatim; `interpolate(0.5)` only checks `not.toThrow()`, which holds on the old regex too.

Assert the interpolated value instead of the endpoint so it fails on the old regex:

```ts
expect(interpolate(0.5)).toBe('translateY(4.75px)')
```

joshuaellis (approved):

